PR #1, opened 2021-05-04
Improve ExternalLinks and make it a service

lucaswerkmeister:


includes/ServiceWiring.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\ExternalLinks;
+use MediaWiki\MediaWikiServices;
+
+/** @phpcs-require-sorted-array */
```
toban, line 6: nice!

.phpcs.xml
```diff
@@ -2,5 +2,10 @@
 <ruleset>
 	<rule ref="./vendor/mediawiki/mediawiki-codesniffer/MediaWiki">
 	</rule>
+
+	<rule ref="MediaWiki.Commenting.FunctionComment">
```
toban, line 6: Was this to make it pass sniffing for now instead of fixing it? Not sure i get what MediaWiki.Commenting.FunctionComment is referring to exactly.

toban (commented):
looks good to me, we could start adding more tests too i guess, do we want tests for existing code?

.phpcs.xml
lucaswerkmeister: > Ignore the PHPCS FunctionComment sniff, since functions can be fully
self-documenting using static parameter and return types.

In other words, PHPCS wants a comment like

```php
/*
 * @param ContainerInterface|null $services
 * @return ExternalLinks
 */
public static function getExternalLinks( ContainerInterface $services = null ): ExternalLinks {
```

and in my opinion that’s just pointless. (We ignore the sniff in Wikibase for mostly the same reason.)

includes/ServiceWiring.php
lucaswerkmeister: It’s even autofixable by PHPCBF 😊

toban: zomg!

.phpcs.xml
toban: yeah that sounds a bit too strict. thanks for explaining!

PR #2, opened 2021-05-05
add workflow file

toban:


addshore:
There are a couple of in progress patches (2 merged, 1 pending) that cleanup these files a little bit form Wikibase.git

https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/684790
https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/684789
https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikibase/+/684788

build/ci-scripts/install.sh
```diff
@@ -0,0 +1,41 @@
+#! /bin/bash
+
+set -x
+
+originalDirectory=$(pwd)
+
+cd ..
+
+mkdir mediawiki
+wget -O- https://github.com/wikimedia/mediawiki/archive/$MW_BRANCH.tar.gz | tar -zxf - -C mediawiki --strip-components 1
```
lucaswerkmeister: `-f -` is redundant in `tar`, `tar -zx` would work just as well :)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
+name: WikibaseReconcileEdit CI
+
+on:
+    push:
```
lucaswerkmeister, line 4: IIRC this should also include `pr:`, to run CI on pull requests that don’t come from this repo?

lucaswerkmeister (commented):
This feels a bit overly complicated to me, to be honest, with things like the `WB` variable that’s only ever set to `"both"` even though we still have code supporting other configurations… not sure how I feel about it.

build/ci-scripts/script.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+set -ex
+
+cd ..
+echo $PWD
+php mediawiki/tests/phpunit/phpunit.php mediawiki/extensions/WikibaseReconcileEdit/tests
```
lucaswerkmeister: We could consider splitting this into two kinds of tests: MediaWiki integration tests, which we run using `mediawiki/tests/phpunit/phpunit.php`, and unit tests, which we run using `mediawiki/vendor/bin/phpunit`. (Compare [T96458](https://phabricator.wikimedia.org/T96458).)

build/ci-scripts/install.sh
toban: > This feels a bit overly complicated to me, to be honest, with things like the `WB` variable that’s only ever set to `"both"` even though we still have code supporting other configurations… not sure how I feel about it.

Well it was a copy and paste job that gave us a lot of flexibility without any real effort. Sure we might not have the need to flip client on or off now but the possibility is there. I guess we could spend more time to trim it down to the bare minimum of what we need right now but that seems ... like a waste of time?

.github/workflows/ci.yml
toban: done

build/ci-scripts/install.sh
toban: > `-f -` is redundant in `tar`, `tar -zx` would work just as well :)

done

build/ci-scripts/script.sh
toban: Yeah sounds good, i guess we need to write one first :)

PR #3, opened 2021-05-05
add COPYING

toban:


LICENSE
```diff
@@ -0,0 +1,348 @@
+The license text below "----" applies to all files within this distribution, other
+than those that are in a directory which contains files named "LICENSE" or
+"COPYING", or a subdirectory thereof. For those files, the license text contained in
+said file overrides any license information contained in directories of smaller depth.
+Alternative licenses are typically used for software that is provided by external
+parties, and merely packaged with the Wikibase release for convenience.
```
lucaswerkmeister: Do we need this part? It talks about “the Wikibase release”, but this isn’t the Wikibase extension…

toban: na looks special lets remove it

addshore (approved):

lucaswerkmeister:
On codesearch it looks like most Wikimedia-deployed extensions use `COPYING` rather than `LICENSE` (the whole BlueSpice suite seems to use both?)… any objections to renaming the file?

LGTM otherwise.

toban:
> On codesearch it looks like most Wikimedia-deployed extensions use `COPYING` rather than `LICENSE` (the whole BlueSpice suite seems to use both?)… any objections to renaming the file?
> 
> LGTM otherwise.

good catch yeah that looks like it would make sense!

PR #5, opened 2021-05-06
Start replacing die() with proper error handling

lucaswerkmeister:
Bug: [T281804](https://phabricator.wikimedia.org/T281804)

---

Note: branch includes #6.

### DIFF
--- a/includes/MediaWiki/Api/EditEndpoint.php
+++ b/includes/MediaWiki/Api/EditEndpoint.php
@@ -63,11 +63,17 @@ class EditEndpoint extends SimpleHandler {
 				400
 			);
 		}
+		$supportedReconciliationVersions = [ '0.0.1' ];
 		if (
 			!array_key_exists( self::VERSION_KEY, $inputReconcile ) ||
-			$inputReconcile[self::VERSION_KEY] !== '0.0.1'
+			!in_array( $inputReconcile[self::VERSION_KEY], $supportedReconciliationVersions )
 		) {
-			die( 'Only supported reconciliation version is 0.0.1' );
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-unsupported-reconcile-version' )
+					->textListParams( $supportedReconciliationVersions )
+					->numParams( count( $supportedReconciliationVersions ) ),
+				400
+			);
 		}
 		if (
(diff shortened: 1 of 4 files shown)


lucaswerkmeister:
I think we’ve reached a good mid-way point now – some initial `die()`s are converted, and the rest are going to require more MediaWiki / Wikibase interaction (any `die()`s past this point require the test to at least set up a property). I’d say we can leave those for later PRs.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -55,19 +57,32 @@ public function run() {
...
+			);
 		}
 		if (
 			!array_key_exists( 'urlReconcile', $inputReconcile ) ||
 			!preg_match( PropertyId::PATTERN, $inputReconcile['urlReconcile'] )
 		) {
-			die( '0.0.1 requires a single urlReconcile key mapped to a property id, such as P123' );
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid' )
+					->textParams( '0.0.1', 'urlReconcile' ),
```
toban: Could we use $inputReconcile[self::VERSION_KEY] here already as we have validated it?

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -55,19 +57,32 @@ public function run() {
 		// TODO output an object that controls the reconciliations spec?
 		$inputReconcile = $request->reconcile();
 		if ( $inputReconcile === null ) {
-			die( 'Invalid reconcile JSON supplied' );
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-json' ),
+				400
+			);
 		}
+		$supportedReconciliationVersions = [ '0.0.1' ];
```
toban, line 66: Not really a part of this task but i could see this moving to some other place soon.

toban (approved):
LGTM just some thoughts

includes/MediaWiki/Api/EditEndpoint.php
lucaswerkmeister: Sure, makes sense.

PR #10, opened 2021-05-07
Compare el_to in ExternalLinks

lucaswerkmeister:
The `el_index` field does not contain all the information in the URL – for instance, it discards the username and password (if the URL contains any). This means that, while we can use it to speed up the query, we still need to compare the `el_to` field as well in order to ensure that we only include page IDs for the right URL.

Bug: [T282650](https://phabricator.wikimedia.org/T282650)

### DIFF
--- a/includes/MediaWiki/ExternalLinks.php
+++ b/includes/MediaWiki/ExternalLinks.php
@@ -28,6 +28,7 @@ class ExternalLinks {
 			'externallinks',
 			'el_from',
 			[
+				'el_to' => $url,
 				'el_index' => $this->getSingleIndexOrDie( $url ),
 			],
 			__METHOD__
--- a/tests/integration/MediaWiki/ExternalLinksTest.php
+++ b/tests/integration/MediaWiki/ExternalLinksTest.php
@@ -70,6 +70,22 @@ class ExternalLinksTest extends \MediaWikiIntegrationTestCase {
 		$this->assertSame( [ '1' ], $result );
 	}
 
+	public function testPageIdsContainingLinksWithDifferentUserinfo(): void {
+		$userinfoAndPageIds = [
+			[ 'me:mypassword', 1 ],
+			[ 'someoneelse:theirpassword', 2 ],
+		];
+		foreach ( $userinfoAndPageIds as [ $userinfo, $pageId ] ) {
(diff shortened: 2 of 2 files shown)


includes/MediaWiki/ExternalLinks.php
```diff
@@ -28,6 +28,7 @@ public function pageIdsContainingUrl( string $url ) : array {
 			'externallinks',
 			'el_from',
 			[
+				'el_to' => $url,
```
toban, line 31: Would we need them both still? This condition would be and AND/OR?

tests/integration/MediaWiki/ExternalLinksTest.php
```diff
@@ -70,6 +70,22 @@ public function testPageIdsContainingALink() {
...
+			[ 'me:mypassword', 1 ],
+			[ 'someoneelse:theirpassword', 2 ],
+		];
+		foreach ( $userinfoAndPageIds as [ $userinfo, $pageId ] ) {
+			$rows = $this->getExternalLinksRows( "http://$userinfo@example.com/", $pageId );
+			$this->db->insert( 'externallinks', $rows, __METHOD__ );
+		}
+
+		$externalLinks = $this->newExternalLinks();
+		$result = $externalLinks->pageIdsContainingUrl( 'http://me:mypassword@example.com/' );
```
toban, line 84: Should we change name on this now as the function now requires an exact match?

```
pageIdsMatchingUrl
```

maybe?

toban (commented):
I'm not sure I completely see the use-case where this would happen in the context of our crawler but looks like a reasonable thing to use instead. We probably only want exact matches here either way?

includes/MediaWiki/ExternalLinks.php
lucaswerkmeister: Yes, we need them both. `el_to` alone can’t be efficiently searched for, that’s why `el_index` exists.

tests/integration/MediaWiki/ExternalLinksTest.php
lucaswerkmeister: I think it was always meant to be an exact match – “containing” refers to the page having more content than just the one URL (and a page can contain multiple URLs).

PR #11, opened 2021-05-07
Inject Wikibase services into EditEndpoint

lucaswerkmeister:
On REL1_35, they’re not yet available in the MediaWiki service container, but getting them in the `factory()` method and injecting them into the constructor is still better than getting them in the `run()` method. One thing this lets us do is inject a fake property datatype lookup in the test, which is probably more efficient than creating and saving a real property.

We pass the `PropertyDataTypeLookup` into the `MinimalItemInput` constructor, but apart from that the item input constructors still get their services from the default `WikibaseRepo` instance. I assume we’ll turn these into services of their own later.

Bug: [T282244](https://phabricator.wikimedia.org/T282244)

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -19,6 +19,7 @@
 			"path": "/wikibase-reconcile-edit/v0/edit",
 			"method": "POST",
 			"class": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Api\\EditEndpoint",
+			"factory": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Api\\EditEndpoint::factory",
 			"services": [
 				"WikibaseReconcileEdit.ExternalLinks"
 			]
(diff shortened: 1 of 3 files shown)


lucaswerkmeister:
> We pass the `PropertyDataTypeLookup` into the `MinimalItemInput` constructor, but apart from that the item input constructors still get their services from the default `WikibaseRepo` instance. I assume we’ll turn these into services of their own later.

@toban I’d say that’s part of https://phabricator.wikimedia.org/T282243? :)

lucaswerkmeister:
Wait a second. We’re only injecting our fake property datatype lookup into `EditEndpoint`, not into Wikibase itself… why can the item still be saved successfully?

lucaswerkmeister:
I guess `EditEntity` gets an already deserialized entity and just assumes that all the referenced entities also exist… in normal API usage, an error that a property can’t be found would’ve been thrown at an earlier point, so it’s not `EditEntity`’s responsibility? And that’s why it works?

PR #14, opened 2021-05-10
Split reconciliation logic to it's own service

toban:
- Splits out the reconciliation part of EditEndpoint to it's own service
- Adds some unit tests
- Currently never hitting ->die which is being refactored out.

Bug: T282243

### DIFF
--- a/includes/MediaWiki/ReconciliationItem.php
+++ b/includes/MediaWiki/ReconciliationItem.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki;
+
+use Wikibase\DataModel\Entity\Item;
+
+/**
+ * Class that contains the item and revision found by the reconciliation service
+ * @see ReconciliationService
+ */
+class ReconciliationItem {
+
+	/** @var Item */
+	private $item;
+
+	/** @var int|bool */
+	private $revision;
+
+	public function __construct( Item $item, $revision ) {
+		$this->item = $item;
+		$this->revision = $revision;
+	}
+
+	/**
+	 * @return Item
+	 */
+	public function getItem(): Item {
+		return $this->item;
+	}
+
+	/**
+	 * @return bool|int
+	 */
+	public function getRevision() {
+		return $this->revision;
+	}
+
+}
--- a/includes/MediaWiki/ReconciliationService.php
+++ b/includes/MediaWiki/ReconciliationService.php
@@ -57,11 +57,6 @@ class ReconciliationService {
 	 */
 	private function getItemIdsFromPageIds( array $pageIds ) : array {
 		$titles = $this->titleFactory->newFromIDs( $pageIds );
-
-		if ( empty( $titles ) ) {
-			return [];
-		}
-
 		$entityIds = $this->entityIdLookup->getEntityIds( $titles );
 		$itemIds = [];
 		foreach ( $entityIds as $entityId ) {
--- a/includes/MediaWiki/ReconciliationService.php
+++ b/includes/MediaWiki/ReconciliationService.php
@@ -81,7 +81,7 @@ class ReconciliationService {
 	 * @return ReconciliationItem
 	 * @throws \Exception
 	 */
-	public function getItemByStatementUrl(
+	public function getOrCreateItemByStatementUrl(
 		PropertyID $reconcileUrlProperty,
 		string $reconciliationUrl
 	) : ReconciliationItem {
--- a/includes/MediaWiki/ReconciliationService.php
+++ b/includes/MediaWiki/ReconciliationService.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki;
+
+use TitleFactory;
+use Wikibase\DataModel\Entity\Item;
+use Wikibase\DataModel\Entity\ItemId;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Lookup\EntityLookup;
+use Wikibase\DataModel\Snak\PropertyValueSnak;
+use Wikibase\Lib\Store\EntityIdLookup;
+use Wikibase\Lib\Store\EntityRevisionLookup;
+use Wikibase\Repo\Store\IdGenerator;
+
+/**
+ * Service for reconciling items based on url statements
+ */
+class ReconciliationService {
+
+	/** @var EntityIdLookup */
+	private $entityIdLookup;
+
+	/** @var EntityLookup */
+	private $entityLookup;
+
+	/** @var EntityRevisionLookup */
+	private $entityRevisionLookup;
+
+	/** @var IdGenerator */
+	private $idGenerator;
+
+	/** @var ExternalLinks */
+	private $externalLinks;
+
+	/** @var TitleFactory */
+	private $titleFactory;
+
+	public function __construct(
+		EntityIdLookup $entityIdLookup,
+		EntityLookup $entityLookup,
+		EntityRevisionLookup $entityRevisionLookup,
+		IdGenerator $idGenerator,
+		ExternalLinks $externalLinks,
+		TitleFactory $titleFactory
+	) {
+		$this->entityIdLookup = $entityIdLookup;
+		$this->entityLookup = $entityLookup;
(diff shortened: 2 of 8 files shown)


toban:
needs a rebase

includes/ServiceWiring.php
```diff
@@ -12,4 +14,16 @@
...
 
+	'WikibaseReconcileEdit.ReconciliationService' => function ( MediaWikiServices $services ): ReconciliationService {
+		$repo = WikibaseRepo::getDefaultInstance();
+
+		return new ReconciliationService(
+			$repo->getEntityIdLookup(),
+			$repo->getEntityLookup(),
+			$repo->getEntityRevisionLookup(),
+			$repo->newIdGenerator(),
+			$services->get( 'WikibaseReconcileEdit.ExternalLinks' )
```
lucaswerkmeister: We have a convenience method for that, `WikibaseReconcileEditServices::getExternalLinks( $services )`.

includes/ServiceWiring.php
```diff
@@ -12,4 +14,16 @@
 		);
 	},
 
+	'WikibaseReconcileEdit.ReconciliationService' => function ( MediaWikiServices $services ): ReconciliationService {
```
lucaswerkmeister, line 38: This new service should also have an accessor method in `WikibaseReconcileEditServices`.

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -46,13 +47,15 @@ private function newHandler() {
...
 		return new EditEndpoint(
 			$repo->newEditEntityFactory(),
-			$repo->getEntityIdLookup(),
-			$repo->getEntityLookup(),
-			$repo->getEntityRevisionLookup(),
-			$repo->newIdGenerator(),
 			$propertyDataTypeLookup,
-			new ExternalLinks(
-				LoadBalancerSingle::newFromConnection( $this->db )
+			new ReconciliationService(
```
lucaswerkmeister: We could probably replace this with `WikibaseReconcileEditServices::getReconciliationService()`, once that method exists – since we’re not injecting any special services in here. (That would replace the `LoadBalancerSingle` with whatever real load balancer is in the service container, but I assume that won’t make a difference in practice.)

tests/unit/MediaWiki/ReconciliationServiceTest.php
```diff
@@ -0,0 +1,185 @@
...
+	public function testBasicConstruction() {
+		$service = new ReconciliationService(
+			$this->createMock( EntityIdLookup::class ),
+			$this->createMock( EntityLookup::class ),
+			$this->createMock( EntityRevisionLookup::class ),
+			WikibaseRepo::getDefaultInstance()->newIdGenerator(),
+			$this->createMock( ExternalLinks::class ),
+		);
+
+		$this->assertInstanceOf( ReconciliationService::class, $service );
```
lucaswerkmeister: This assertion is guaranteed to succeed, since we literally created the object with `new ReconciliationService()`… I’m not sure what the point of this test is, to be honest.

includes/MediaWiki/ReconciliationItem.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki;
+
+use Wikibase\DataModel\Entity\Item;
+
+/**
+ * Class that contains the item and revision found by the reconciliation service
+ * @see ReconciliationService
+ */
+class ReconciliationItem {
```
lucaswerkmeister, line 11: Maybe we could use the Wikibase `EntityRevision` class instead of this? (Though in that case the content would only be known to be an entity, not specifically an item.)

includes/MediaWiki/ReconciliationService.php
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Finds existing item or returns a new one if no matching item is found.
+	 * Lookup is matching on items with statement ($reconcileUrlProperty) set to the $reconciliationUrl
+	 *
+	 * @param PropertyId $reconcileUrlProperty
+	 * @param string $reconciliationUrl
+	 * @return ReconciliationItem
+	 * @throws \Exception
+	 */
+	public function getItemByStatementUrl(
```
lucaswerkmeister: What do you think about changing “get” to “getOrCreate” here, to make it clearer that the method may create a new item?

tests/unit/MediaWiki/ReconciliationServiceTest.php
toban: grand, lets remove it

includes/MediaWiki/ReconciliationService.php
toban: yeah sounds good

includes/ServiceWiring.php
toban: done

includes/ServiceWiring.php
toban: done

includes/MediaWiki/ReconciliationService.php
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/**
+	 * @param int[] $pageIds
+	 * @return ItemId[]
+	 */
+	private function getItemIdsFromPageIds( array $pageIds ) : array {
+		$titles = $this->titleFactory->newFromIDs( $pageIds );
+
+		if ( empty( $titles ) ) {
```
lucaswerkmeister: Is this added condition necessary? Do you know if anything breaks if we call `EntityIdLookup::getEntityIds()` with an empty array?

includes/MediaWiki/ReconciliationService.php
```diff
@@ -0,0 +1,144 @@
...
+
+		// If we have more than one item matches, something is wrong and we can't edit
+		if ( count( $itemsThatReferenceTheUrlInCorrectStatement ) > 1 ) {
+			die( 'Matched multiple Items during reconciliation :(' );
+		}
+
+		// Get our base
+		if ( count( $itemsThatReferenceTheUrlInCorrectStatement ) === 1 ) {
+			$base = $itemsThatReferenceTheUrlInCorrectStatement[0];
+			// XXX: This bit is so annoying...
```
lucaswerkmeister, line 116: We can probably alleviate this by using an `EntityRevisionLookup` rather than an `EntityLookup` in the earlier block where we load the item, though that should be done in a separate commit.

includes/MediaWiki/ReconciliationItem.php
toban: Which would mean we would have to keep checking for type and cast it (or refactor all other things relying on this being an item). 

Do we see a use-case for anything other than items? I guess we could extend EntityRevision with some helper function that casts the item but that feels kinda awkward too considering how simple both classes are.

toban:
```

/**
 * Class that contains the item and revision found by the reconciliation service
 * @see ReconciliationService
 */
class ReconciliationItem extends EntityRevision {

	public function __construct( Item $item, $revision ) {
		parent::__construct( $item, $revision );
	}

	/**
	 * @return Item
	 */
	public function getItem(): Item {
		/** @var Item $item */
		$item = $this->getEntity();
		return $item;
	}
}
```
¯\_(ツ)_/¯

includes/MediaWiki/ReconciliationService.php
toban: If not then I had to add a bunch of mocks for the functions these services use in the test. Which just makes the tests harder to read and more work than exiting early here which should be fine.

includes/MediaWiki/ReconciliationItem.php
lucaswerkmeister: We could even just do

```php
/** @var Item $item */
$item = $entityRevision->getEntity();
```

and call it a day, since PHP is a dynamic language…

toban: After a chat, lets keep the thing as it is and maybe refactor this later.

includes/MediaWiki/ReconciliationService.php
toban: lets get rid of it

includes/MediaWiki/ReconciliationService.php
toban: put this in it's own task https://phabricator.wikimedia.org/T282553 so we don't forget

PR #18, opened 2021-05-12
Replace die() with proper error reporting

guergana:
Continuing replacing dies with LocalizedHttpExceptions.

Will do separate PRs for ease of reviewing.

Bug: T281804

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -13,5 +13,6 @@
 	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Entity input version must be specified",
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
-	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property"
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must be of type value"
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -8,5 +8,9 @@
 	"wikibasereconcileedit-desc": "This is the ON OR Prototype extension",
 	"wikibasereconcileedit-editendpoint-invalid-reconcile-json": "Invalid reconcile JSON supplied",
 	"wikibasereconcileedit-editendpoint-unsupported-reconcile-version": "The only supported reconciliation {{PLURAL:$2|version is|versions are}} $1",
-	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123"
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
+	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Entity input version must be specified",
+	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
+	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported"
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -12,5 +12,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
 	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Entity input version must be specified",
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
-	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported"
+	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -12,5 +12,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-type-property-must-be-url": "Error returned by the API if the “reconcile” request propertyId parameter is not of type URL.\n\nParameters:\n* $1 – the name of the property, e.g. P123\n* $2 – the name of the field, e.g. urlReconcile.\n* $3 – the datatype of the provided reconcile property, e.g. string",
 	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Error returned by the API if the “entity” request parameter contains no version field in the entity input",
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions\n",
-	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported."
+	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -13,5 +13,6 @@
 	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Error returned by the API if the “entity” request parameter contains no version field in the entity input",
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions\n",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
-	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property"
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -8,5 +8,9 @@
 	"wikibasereconcileedit-desc": "{{desc|name=WikibaseReconcileEdit}}",
 	"wikibasereconcileedit-editendpoint-invalid-reconcile-json": "Error returned by the API if the “reconcile” request parameter is invalid.",
 	"wikibasereconcileedit-editendpoint-unsupported-reconcile-version": "Error returned by the API if the “reconcile” request parameter specifies an unsupported version.\n\nParameters:\n* $1 – the supported versions (comma-separated list)\n* $2 – the number of supported versions",
-	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "Error returned by the API if the “reconcile” request parameter contains no valid field to specify the property ID to reconcile on.\n\nParameters:\n* $1 – the version of the request parameter, e.g. 0.0.1\n* $2 – the name of the field, e.g. urlReconcile"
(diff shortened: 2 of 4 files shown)


includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -137,27 +137,40 @@ public function run() {
 		$reconcileUrlProperty = new PropertyId( $inputReconcile['urlReconcile'] );
 		// For now this property must be of URL type
 		if ( $this->propertyDataTypeLookup->getDataTypeIdForProperty( $reconcileUrlProperty ) !== 'url' ) {
-			die( 'urlReconcile property must be of type url' );
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-type-property-must-be-url' ),
```
lucaswerkmeister: I think it would be useful to include the actual datatype as a message argument here.

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -26,7 +26,9 @@ class EditEndpointTest extends \MediaWikiIntegrationTestCase {
 	use HandlerTestTrait;
 
 	private const URL_PROPERTY = 'P1';
+	private const STRING_PROPERTY = 'P2';
 	private const MISSING_PROPERTY = 'P1000';
+	private const QUALIFIER_PROPERTY = 'P666';
```
lucaswerkmeister: I find this confusing – qualifiers use the same properties in Wikibase as everything else, and “qualifier” is not a datatype in the same way that “url” and “string” are. I suggest removing this, and using `STRING_PROPERTY` in the tests below.

i18n/en.json
```diff
@@ -8,5 +8,9 @@
 	"wikibasereconcileedit-desc": "This is the ON OR Prototype extension",
 	"wikibasereconcileedit-editendpoint-invalid-reconcile-json": "Invalid reconcile JSON supplied",
 	"wikibasereconcileedit-editendpoint-unsupported-reconcile-version": "The only supported reconciliation {{PLURAL:$2|version is|versions are}} $1",
-	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123"
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "urlReconcile property must be of type url",
+	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "entity input version must be specified",
```
lucaswerkmeister: I would use the opportunity in this pull request to at least begin this and the next message with an uppercase character (“Entity”, “Unknown”), to have a little bit of consistency in the messages.

i18n/en.json
```diff
@@ -8,5 +8,9 @@
 	"wikibasereconcileedit-desc": "This is the ON OR Prototype extension",
 	"wikibasereconcileedit-editendpoint-invalid-reconcile-json": "Invalid reconcile JSON supplied",
 	"wikibasereconcileedit-editendpoint-unsupported-reconcile-version": "The only supported reconciliation {{PLURAL:$2|version is|versions are}} $1",
-	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123"
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "urlReconcile property must be of type url",
```
lucaswerkmeister: The urlReconcile name should probably be a message parameter, like $2 in the previous message.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -137,27 +137,40 @@ public function run() {
...
 		$inputEntityVersion = $request->entity()[self::VERSION_KEY];
 		if ( $inputEntityVersion === '0.0.1/full' ) {
 			$inputEntity = ( new FullWikibaseItemInput )->getItem( $request );
 		} elseif ( $inputEntityVersion === '0.0.1/minimal' ) {
 			$inputEntity = ( new MinimalItemInput( $this->propertyDataTypeLookup ) )->getItem( $request );
 		} else {
-			die( 'unknown entity input version' );
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-entity-input-version' )
+					->textListParams( [ '0.0.1/full', '0.0.1/minimal' ] ),
```
lucaswerkmeister: This parameter isn’t currently used in the message. If we want to include it, I think we should also include the `$inputEntityVersion`, and we’ll also need the number of allowed versions for pluralization (like in an earlier message). So the message text could be something like:

> Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2.

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -400,6 +400,44 @@ public function testReconciliationPropertyMissingInStatements(): void {
...
+				'entity' => [
+					EditEndpoint::VERSION_KEY => '0.0.1/full',
+					'type' => 'item',
+					'claims' => [
+						self::URL_PROPERTY => [
+							[
+								'mainsnak' => [
+									'snaktype' => 'somevalue',
+									'property' => self::URL_PROPERTY,
+									'datavalue' => [
```
lucaswerkmeister: somevalue snaks should not have a datavalue.

i18n/en.json
```diff
@@ -13,5 +13,6 @@
 	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "entity input version must be specified",
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "unknown entity input version",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
-	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property"
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must be of type value"
```
lucaswerkmeister, line 17: “must be of type value” sounds pretty technical to me; maybe “must contain a value” would be clearer?

i18n/en.json
```diff
@@ -14,5 +14,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "unknown entity input version",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must be of type value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must be of type value",
+	"wikibasereconcileedit-editendpoint-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\nThe matched items are: $1\n"
```
lucaswerkmeister: I feel like this should use `{{PLURAL}}`. In English, it will never select the singular version of the message, but some other languages have other distinctions than “one” / “other”, and including `{{PLURAL}}` in the English source message makes it clearer for translators that they can also use it in translations, I think. But I’m not sure how this is handled in other messages.

i18n/en.json
guergana: Done.

i18n/en.json
guergana: Done

includes/MediaWiki/Api/EditEndpoint.php
guergana: You are right. :sweat_smile: I didn't use the parameter. Now done. :)

tests/integration/MediaWiki/Api/EditEndpointTest.php
guergana: I thought using the String property would make other tests fail because of the datatype is not URL. But the tests are passing.

i18n/en.json
guergana: Done.

i18n/en.json
guergana: Done.

i18n/en.json
guergana: Makes sense. Done.

includes/MediaWiki/Api/EditEndpoint.php
guergana: Agree. Done.

includes/MediaWiki/Api/EditEndpoint.php
guergana: True, I didn't add it :sweat_smile: Done.

tests/integration/MediaWiki/Api/EditEndpointTest.php
guergana: I thought some of the tests would fail if I compare them to a property with a datatype that's not URL. But they are passing. Done!

i18n/en.json
guergana: Done.

tests/integration/MediaWiki/Api/EditEndpointTest.php
guergana: Done. thanks I didn't know this.

i18n/qqq.json
```diff
@@ -8,5 +8,12 @@
...
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "Error returned by the API if the “reconcile” request parameter contains no valid field to specify the property ID to reconcile on.\n\nParameters:\n* $1 – the version of the request parameter, e.g. 0.0.1\n* $2 – the name of the field, e.g. urlReconcile",
+	"wikibasereconcileedit-editendpoint-invalid-type-property-must-be-url": "Error returned by the API if the “reconcile” request propertyId parameter is not of type URL.\n\nParameters:\n* $1 – the name of the property, e.g. P123\n* $2 – the name of the field, e.g. urlReconcile.\n* $3 – the datatype of the provided reconcile property, e.g. string",
+	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Error returned by the API if the “entity” request parameter contains no version field in the entity input",
+	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 – the supported versions (comma-separated list)\n",
```
lucaswerkmeister: The parameters here don’t look right – in the English message, $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions.

i18n/en.json
```diff
@@ -8,5 +8,12 @@
...
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
+	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Entity input version must be specified",
+	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
+	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-editendpoint-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\n{{PLURAL:$2|version is|versions are}} $1\n"
```
lucaswerkmeister: Wait, why “version”?

guergana: Because copy/paste :facepalm: Sorry, this is my first time using PLURAL.

i18n/qqq.json
guergana: You are right. Updated!

i18n/en.json
```diff
@@ -8,5 +8,12 @@
...
+	"wikibasereconcileedit-editendpoint-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-editendpoint-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
+	"wikibasereconcileedit-editendpoint-unspecified-entity-input-version": "Entity input version must be specified",
+	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
+	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
+	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-editendpoint-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\n{{PLURAL:$2|item is|items are}} $1\n"
```
lucaswerkmeister: Hang on, this message isn’t actually used (probably because the corresponding `die()` was moved to `ReconciliationService` in the meantime). I guess we should remove the last commit for now?

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -119,38 +119,62 @@ public function run() {
...
+				throw new LocalizedHttpException(
+					MessageValue::new( 'wikibasereconcileedit-editendpoint-qualifiers-references-not-supported' ),
+					400
+				);
 			}
 		}
 		// Check for our reconciliation value
 		$reconciliationStatements = $inputEntity->getStatements()->getByPropertyId( $reconcileUrlProperty );
 		if ( $reconciliationStatements->count() !== 1 ) {
-			die( 'Entity must have at least one statement for the reconciliation Property' );
```
lucaswerkmeister, line 149: I notice now that this error message is a bit inaccurate – it happens not just when there are no statements for the reconciliation property, but also when there are multiple (compare `testMatchedMultipleItems()` below). Should we fix that now or later?

i18n/en.json
guergana: :(

includes/MediaWiki/Api/EditEndpoint.php
guergana: Later. I would like to merge this sooner than later.

guergana: Should we update the logic or the message would be enough?

lucaswerkmeister: I think updating the logic would be nicer, with separate messages.

lucaswerkmeister (approved):

PR #17, opened 2021-05-11
Reconcile item statement values

lucaswerkmeister:
Adds reconciliation for `wikibase-item` statements through the reconciliation url

PR #19, opened 2021-05-12
Stop clearing wb_property_info table in EditEndpointTest

lucaswerkmeister:
Follow-up to fbb0043c7b (#11): since we inject a fake property datatype lookup instead of creating a real property, we no longer need to wipe the property info table between tests.

Bug: [T282244](https://phabricator.wikimedia.org/T282244)


lucaswerkmeister (commented):
<abbr title="do not merge">DNM</abbr> – we may actually want to create real properties after all (depending on what happens in #17).

lucaswerkmeister:
We’re creating real properties again.

PR #21, opened 2021-05-12
Use only EntityRevisionLookup in ReconciliationService

lucaswerkmeister:
This lets us get rid of the very ugly code to consume the `getLatestRevisionId()` result, and also ensures the item content and revision ID belong together (since they now come from the same source).

Bug: [T282553](https://phabricator.wikimedia.org/T282553)


toban:
nice!

PR #22, opened 2021-05-12
Continuation on #17

toban:
- add a integration test
- insert properties again

### DIFF
--- a/includes/InputToEntity/MinimalItemInput.php
+++ b/includes/InputToEntity/MinimalItemInput.php
@@ -118,6 +118,19 @@ class MinimalItemInput {
 		if ( $name === 'wikibase-item' && wfParseUrl( $value ) !== false ) {
 			$reconciliationItem = $this->reconciliationService
 				->getOrCreateItemByStatementUrl( $reconcileUrlProperty, $value );
+
+			// if this is a new item, we need to put the reconciliation statement in it
+			if ( !$reconciliationItem->getRevision() ) {
+				$parser = $this->valueParserFactory->newParser( 'url', new ParserOptions );
+				$parseResult = $parser->parse( $value );
+				if ( !$parseResult instanceof DataValue ) {
+					die( 'Failed to parse statement value' );
+				}
+				$reconciliationItem->getItem()->getStatements()->addNewStatement(
+					new PropertyValueSnak( $reconcileUrlProperty, $parseResult )
+				);
+			}
+
 			return [ new EntityIdValue( $reconciliationItem->getItem()->getId() ), [ $reconciliationItem ] ];
 		}
 		// TODO add specific options?
--- a/includes/InputToEntity/MinimalItemInput.php
+++ b/includes/InputToEntity/MinimalItemInput.php
@@ -3,6 +3,7 @@
 namespace MediaWiki\Extension\WikibaseReconcileEdit\InputToEntity;
 
 use DataValues\DataValue;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\ReconciliationItem;
 use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\ReconciliationService;
 use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequest;
 use ValueParsers\ParserOptions;
@@ -43,7 +44,7 @@ class MinimalItemInput {
 
 	/**
 	 * @param EditRequest $request
-	 * @return array( Item, Item[] )
+	 * @return array( Item, ReconciliationItem[] )
 	 */
 	public function getItem( EditRequest $request, PropertyId $reconcileUrlProperty ) : array {
 		$inputEntity = $request->entity();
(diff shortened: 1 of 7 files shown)


includes/InputToEntity/MinimalItemInput.php
```diff
@@ -43,7 +44,7 @@ public function __construct(
 
 	/**
 	 * @param EditRequest $request
-	 * @return array( Item, Item[] )
+	 * @return array( Item, ReconciliationItem[] )
```
toban, line 47: Maybe this could become ReconciledDataValue

includes/InputToEntity/MinimalItemInput.php
```diff
@@ -115,10 +116,22 @@ private function getDataValue(
...
+			$reconciliationItem = $this->reconciliationService
+				->getOrCreateItemByStatementUrl( $reconcileUrlProperty, $value );
+
+			// if this is a new item, we need to put the reconciliation statement in it
+			if ( !$reconciliationItem->getRevision() ) {
+				$parser = $this->valueParserFactory->newParser( 'url', new ParserOptions );
+				$parseResult = $parser->parse( $value );
+				if ( !$parseResult instanceof DataValue ) {
+					die( 'Failed to parse statement value' );
+				}
```
lucaswerkmeister: I think here we could just use `new StringValue( $value )` instead of the whole value parser business?

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -188,11 +189,34 @@ public function run() {
...
+			if ( !( $otherItem instanceof ReconciliationItem ) ) {
+				continue;
+			}
+
+			// don't need to save this again
+			if ( $otherItem->getRevision() ) {
+				continue;
+			}
+
+			// don't need to save references to itself again
```
lucaswerkmeister: I think this comment is a bit misleading… if I understand correctly, the item hasn’t actually been saved yet, right? It will be saved (“again”, unless we skip it here) _later_, when we `attemptSave( $toSave )`.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -188,11 +189,34 @@ public function run() {
...
+				$otherItem->getItem()->getId(),
+				false
+			);
+
+			$saveStatus->merge( $otherItemEdit->attemptSave(
+				$otherItem->getItem(),
 				'Reconciliation Edit',
-				EDIT_NEW, // TODO assert $otherItem->getRevision() === null?
-				false // TODO actually do a token check?
+				// TODO assert $otherItem->getRevision() === null?
```
lucaswerkmeister: This can probably be removed, since we already checked earlier that the `getRevision()` is null.

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -25,34 +28,52 @@ class EditEndpointTest extends \MediaWikiIntegrationTestCase {
 
 	use HandlerTestTrait;
 
-	private const URL_PROPERTY = 'P1';
+	/** @var PropertyId URL_PROPERTY */
+	private $URL_PROPERTY = 'P1';
```
lucaswerkmeister: If this is a class member, not a constant, then it shouldn’t be in all-uppercase in my opinion. (Same for the string property too.)

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -25,34 +28,52 @@ class EditEndpointTest extends \MediaWikiIntegrationTestCase {
...
 		parent::setUp();
 		$this->tablesUsed[] = 'page';
+		$this->tablesUsed[] = 'externallinks';
 		$this->tablesUsed[] = 'wb_property_info';
-	}
+		$this->tablesUsed[] = 'wb_id_counters';
 
-	private function newHandler() {
-		$propertyDataTypeLookup = new InMemoryDataTypeLookup();
+		/** @var PropertyId URL_PROPERTY */
```
lucaswerkmeister: I think this comment can be removed? (And its copy a few lines below as well.)

tests/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -25,34 +28,52 @@ class EditEndpointTest extends \MediaWikiIntegrationTestCase {
...
+		WikibaseRepo::getDefaultInstance()->getEntityStore()->saveEntity(
+			new Property( new PropertyId( $this->URL_PROPERTY ), null, 'url' ),
+			__METHOD__,
+			User::newSystemUser( 'TestUSer' ),
```
lucaswerkmeister: Uppercase S in TestUSer looks accidental?

Also, this should probably use `MediaWikiIntegrationTestCase::getTestUser()->getUser()` anyways.

includes/InputToEntity/MinimalItemInput.php
toban: done in https://github.com/wmde/WikibaseReconcileEdit/pull/22/commits/60cdc423354e5051ad5d37faa3fd3890d1294f9c

includes/MediaWiki/Api/EditEndpoint.php
toban: agreed

includes/MediaWiki/Api/EditEndpoint.php
toban: done!

tests/integration/MediaWiki/Api/EditEndpointTest.php
toban: yup!

tests/integration/MediaWiki/Api/EditEndpointTest.php
toban: thanks

PR #25, opened 2021-05-14
Continue replacement of die with proper error handling

guergana:
* Fix invalid-reconciliation-statement-type message
* Multiple item matches
* Remove 'die' for DataValue Exception
* Minimal Input Statement Missing Property or Value Keys

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -15,5 +15,6 @@
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1"
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key"
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -14,5 +14,5 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must be of type value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value"
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -14,5 +14,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -14,6 +14,7 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions\n",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter of the reconciliation statement's mainsnak is not of type value",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items",
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -14,5 +14,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions\n",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter of the reconciliation statement's mainsnak is not of type value",
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items"
 }
--- a/includes/MediaWiki/Api/EditEndpoint.php
+++ b/includes/MediaWiki/Api/EditEndpoint.php
@@ -4,6 +4,7 @@ namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Api;
 
 use DataValues\StringValue;
 use MediaWiki\Extension\WikibaseReconcileEdit\EditStrategy\SimplePutStrategy;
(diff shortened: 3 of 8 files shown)


lucaswerkmeister (dismissed):

lucaswerkmeister:
Should I merge this already or do you want to add more commits? :)

guergana:
> Should I merge this already or do you want to add more commits? :)

Sorry, should have marked it as draft. More commits coming up but we can also merge and I can make a new one :woman_shrugging:

i18n/qqq.json
```diff
@@ -14,5 +14,7 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Error returned by the API if the “entity” request parameter specifies an unsupported version.\n\nParameters:\n* $1 is the specified (unsupported) version, $2 the supported versions, and $3 the number of supported versions\n",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
```
toban, line 17: request parameter ~~if~~ **of** the

includes/MediaWiki/ReconciliationException.php
```diff
@@ -0,0 +1,25 @@
...
+use Exception;
+use Wikimedia\Message\MessageValue;
+
+/**
+ * Base class for all WikibaseReconcileEdit exceptions
+ *
+ * @license GPL-2.0-or-later
+ */
+class ReconciliationException extends Exception {
+
```
lucaswerkmeister, line 18: The class should also declare `private $messageValue` with a suitable `@var` comment.

i18n/en.json
```diff
@@ -14,5 +14,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value"
+	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\n{{PLURAL:$2|item is|items are}} $1\n"
```
lucaswerkmeister: I think we can remove the emoticon here :D suggestion:

> The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1

includes/MediaWiki/ReconciliationService.php
```diff
@@ -105,7 +106,9 @@ public function getOrCreateItemByStatementUrl(
 
 		// If we have more than one item matches, something is wrong and we can't edit
 		if ( count( $itemsThatReferenceTheUrlInCorrectStatement ) > 1 ) {
-			die( 'Matched multiple Items during reconciliation :(' );
+			throw new ReconciliationException(
+				MessageValue::new( 'wikibasereconcileedit-reconciliationservice-matched-multiple-items' )
```
lucaswerkmeister, line 131: The message arguments (items, number of items) seems to be missing here.

includes/MediaWiki/ReconciliationService.php
```diff
@@ -105,7 +106,9 @@ public function getOrCreateItemByStatementUrl(
 
 		// If we have more than one item matches, something is wrong and we can't edit
 		if ( count( $itemsThatReferenceTheUrlInCorrectStatement ) > 1 ) {
-			die( 'Matched multiple Items during reconciliation :(' );
+			throw new ReconciliationException(
```
lucaswerkmeister, line 130: This exception should probably be caught in `EditEndpoint` (and rethrown as a `LocalizedHttpException` with the same message).

tests/unit/InputToEntity/MinimalItemInputTest.php
```diff
@@ -99,4 +100,72 @@ public function testGetItem( array $requestEntity, Item $expected ) {
 		);
 	}
 
+	public function testStatementsHavePropertyKey() {
+		$requestEntityMissingProperty = [
+			'wikibasereconcileedit-version' => '0.0.1/minimal',
+			'labels' => [ 'en' => 'en-label' ],
```
lucaswerkmeister: We could also leave out the labels, descriptions, aliases, and sitelinks, to keep the test cases smaller.

i18n/en.json
```diff
@@ -15,5 +15,6 @@
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\n{{PLURAL:$2|item is|items are}} $1\n"
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Matched $2 Items during reconciliation :(\n\n{{PLURAL:$2|item is|items are}} $1\n",
+	"wikibasereconcileedit-minimaliteminput-required-property-value-keys": "Statements must have property and value keys"
```
lucaswerkmeister: Maybe the words “property” and “value” should be message parameters so that we can change them without changing the message… what do you think?

Or maybe this should just be one message about a missing key, and then we check the keys separately in the service, and throw the message with either a “property” or “value” parameter?

(If we don’t go with that, then the `qqq.json` should at least mention that the words “property” and “value” should not be translated, in my opinion.)

includes/MediaWiki/ReconciliationException.php
guergana: True! Thanks.

i18n/en.json
guergana: haha, i thought it was cute. But yes i can remove it.

includes/MediaWiki/ReconciliationService.php
guergana: True! :scream:

includes/MediaWiki/ReconciliationService.php
guergana: This was in fact in EditEndPoint before. And was moved to this file as I was working on this commit. Am I understanding right and you are suggesting having it in both places???

tests/unit/InputToEntity/MinimalItemInputTest.php
guergana: True :)

i18n/en.json
guergana: I would go with the third option. It's always going to be either the property or the value, right? The user can probably figure out which one is wrong.

toban (commented):
LGTM just noticed a typo. 

https://github.com/wmde/WikibaseReconcileEdit/pull/25#discussion_r633428198

Should probably also be dealt with as the ReconciliationException would be unhandled and give the API user no helpful messages to debug what went wrong (the page would crash with a 500).

includes/MediaWiki/ReconciliationService.php
lucaswerkmeister: No, I’m saying EditEndpoint should have something like

```php
try {
	// ...
} catch ( ReconciliationException $e ) {
	throw new LocalizedHttpException( $e->getMessageValue(), 400 );
}
```

i18n/qqq.json
guergana: thanks!

includes/MediaWiki/ReconciliationService.php
guergana: Ah, gotcha. Thanks for the example

guergana: ok :)

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -182,10 +183,14 @@ public function run() {
 		$reconciliationDataValue = $reconciliationMainSnak->getDataValue();
 		$reconciliationUrl = $reconciliationDataValue->getValue();
 
-		$reconciliationItem = $this->reconciliationService->getOrCreateItemByStatementUrl(
-			$reconcileUrlProperty,
-			$reconciliationUrl
-		);
+		try {
```
lucaswerkmeister, line 105: The call to `$this->minimalItemInput->getItem()` earlier (ca. line 140) can also throw a `ReconciliationException`, if I’m not mistaken, so I think this try-catch needs to be “widened” quite a bit (or add a separate one up there).

Alternatively, if #26 gets merged first, the try-catch only needs to wrap one line if I’m not mistaken:

```php
$request = $this->editRequestParser->parseRequestInterface( $this->getRequest() );
```

guergana: Done.

i18n/en.json
toban: Hmm, `wikibasereconcileedit-minimaliteminput-required-property-value-keys`seems to have disappeared in the branch is this still an issue?

tests/unit/InputToEntity/MinimalItemInputTest.php
toban: Can't find this test anymore.

i18n/en.json
toban: done

toban:
- Updated the copy of the error message.
- Updated the phpdoc for errors thrown by `ReconcilationService::getOrCreateItemByStatementUrl`

i18n/en.json
guergana: Should be back there now.

tests/unit/InputToEntity/MinimalItemInputTest.php
guergana: Test should be back now. Please re-check :)

i18n/en.json
lucaswerkmeister: ~~The sad smiley is back now?~~

Disregard that, I was looking at an outdated local version of the branch.

tests/unit/InputToEntity/MinimalItemInputTest.php
lucaswerkmeister: ~~Hm, I don’t see any added tests in `MinimalItemInputTest` (only `ReconciliationServiceTest`).~~

Disregard that, I was looking at an outdated local version of the branch.

i18n/en.json
lucaswerkmeister: ~~I think this part got accidentally dropped? As far as I can tell, the `die-error-handling-2` branch currently still doesn’t have this message, and also still has `die( 'statements must have property and value keys' )`.~~

Disregard that, I was looking at an outdated local version of the branch.

tests/unit/InputToEntity/MinimalItemInputTest.php
guergana: Tests are here: https://github.com/wmde/WikibaseReconcileEdit/pull/25/commits/2fa36d7111ba28c906d7b8ecf4d2fc3232b05855#diff-59286771e823077ff1d0248e0248382e9ebdab648bae474dcf31bf61aeb817d3

i18n/en.json
lucaswerkmeister: Okay, but the branch is still missing any of the “three options” as far as I can tell.

I also feel like the third option is needlessly user-hostile. We already know which of the two keys is missing – we even have separate tests for the two cases – so why should we combine them into one, less specific message?

I think the “second option” (one message for a missing key, with “property” or “value” in an argument) is the best one – we can then also use this message later when the keys change (e.g. we add other ones).

i18n/en.json
guergana: Will marked as resolved. This is outdated.

i18n/en.json
guergana: The comment was there before in `qqq.json`. Must have been lost underway. The third option was also suggested by you. :sweat_smile: I re-used the message that was there before but I can also separate them into two different messages of course.

lucaswerkmeister: Yes, but that doesn’t mean I have to like the third option :P

guergana: Please re-check. I've added an if/else. If we add more keys we should probable use a switch.

i18n/qqq.json
```diff
@@ -14,5 +14,7 @@
...
+	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items",
+	"wikibasereconcileedit-minimaliteminput-required-property-value-keys": "Error returned by the API if the minimal item input statements parameter doesn't have property and value keys defined.\n\nParameters:\n $1 - the missing key"
```
lucaswerkmeister: Maybe change “doesn’t have property and value keys defined”? E.g. “is missing a certain key”.

Also, we can remove the `property-value` from the message key. I think just `…-required-keys` should work. (Or maybe `…-required-statements-keys`, because it’s part of the `statements` array.)

guergana: Good catch! Done.

lucaswerkmeister (approved):

PR #30, opened 2021-05-19
Add infrastructure for Cloud VPS test system

lucaswerkmeister:
This adds an Ansible playbook and related files for automatically setting up a WikibaseReconcileEdit test system on Wikimedia Cloud VPS. The setup is largely based on [that of WikibaseManifest][1]; significant differences include:

- We target MediaWiki and Wikibase 1.35, not master. Consequently, Wikibase is loaded using the PHP entry points.
- The instance hostname is changed from *.eqiad.wmflabs to *.eqiad1.wikimedia.cloud, and the domain name from *.wmflabs.org to *.wmcloud.org.
- The wiki does not allow anonymous editing nor registration. Use `createAndPromote.php` to create initial accounts.
- The script path is `/w`, and the article path is `/wiki/$1`; for the script path, `/var/www/html/w` is set up as a symlink to `/var/www/html`, as in WikibaseManifest, but we do it inside `setup.yml` rather than in the `Dockerfile`, because any `/var/www/html` modifications in the `Dockerfile` get lost when docker-compose mounts a different directory over `/var/www/html`.
- Several `tasks` in `setup.yml` are moved to `pre_tasks`; this is necessary to ensure that the `SCRIPTS_DIR` exists by the time the git_updater role runs, which uses that directory.
- The Query Service setup is removed, since we don’t need it and it requires a lot of RAM.
- WikibaseImport is removed, since we don’t need it.

[1]: https://gerrit.wikimedia.org/g/mediawiki/extensions/WikibaseManifest/+/0004d5b7bdf3649cd70c0a57868111277be10cf8/infrastructure/

Bug: [T282230](https://phabricator.wikimedia.org/T282230)


infrastructure/README.md
```diff
@@ -0,0 +1,48 @@
+This provides a quick and automated way to configure a test system for a 'WikibaseReconcileEdit' enabled Wikibase instance
+
+### Prerequisites
```
toban, line 3: Maybe we could add this, i think it's required to reach it?

```
Host wb-reconcile.wikidata-dev.eqiad1.wikimedia.cloud
  HostName wb-reconcile.wikidata-dev.eqiad1.wikimedia.cloud
  User <YOUR_USERNAME>
  ProxyJump bastion.wmflabs.org:22
```

toban (commented):
Tried running it and seems to work as intended. Nice :+1: 

The only thing that we might want to is the `"/wiki` url support (and `w` probably?)

I tried just adding the symlink and this htaccess snippet with a re-deployed, but just ended up breaking it :(

infrastructure/README.md
lucaswerkmeister: Well, I’d expect people to have some general Cloud VPS setup for all hosts, along these lines:

```
Host *.wikimedia.cloud *.wmflabs
User YOURUSER

Host bastion-cloud
HostName bastion.wmcloud.org
User YOURUSER

Host *.eqiad1.wikimedia.cloud *.eqiad.wmflabs
ProxyJump bastion-cloud
```

That should work more generally for all cloud VPS hosts.

lucaswerkmeister:
Pushed a new version with /w + /wiki support.

infrastructure/README.md
toban: nice :+1:

toban (approved):

PR #31, opened 2021-05-19
Last replace die with proper error reporting

guergana:
* Catch fullwikibaseiteminput-unsupported-type exception
* Replace die in getSingleIndexOrThrowException

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -16,5 +16,6 @@
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key"
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
+	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1."
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -16,5 +16,6 @@
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key"
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key",
+	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "Error returned by the API if the full wikibase item input's type parameter is not supported.\n\nParameters:\n* $1 – the supported types (comma-separated list)\n* $2 – the number of supported types"
 }
--- a/includes/InputToEntity/FullWikibaseItemInput.php
+++ b/includes/InputToEntity/FullWikibaseItemInput.php
@@ -3,7 +3,9 @@
 namespace MediaWiki\Extension\WikibaseReconcileEdit\InputToEntity;
 
 use Deserializers\Deserializer;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\ReconciliationException;
 use Wikibase\DataModel\Entity\Item;
+use Wikimedia\Message\MessageValue;
 
 class FullWikibaseItemInput {
 
@@ -20,10 +22,14 @@ class FullWikibaseItemInput {
 	}
 
 	public function getItem( array $inputEntity ) : Item {
-		if ( !array_key_exists( 'type', $inputEntity ) || $inputEntity['type'] !== 'item' ) {
-			die( 'Only supported entity type is \'item\'' );
+		$supportedTypes = [ 'item' ];
+		if ( !array_key_exists( 'type', $inputEntity ) || !in_array( $inputEntity['type'], $supportedTypes ) ) {
+			throw new ReconciliationException(
+				MessageValue::new( 'wikibasereconcileedit-fullwikibaseiteminput-unsupported-type' )
+					->textListParams( $supportedTypes )
+					->numParams( count( $supportedTypes ) )
+			);
 		}
-
 		return $this->deserializer->deserialize( $inputEntity );
 	}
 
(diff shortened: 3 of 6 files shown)


i18n/en.json
```diff
@@ -16,5 +16,6 @@
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key"
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
+	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "Only supported entity type is '$1'"
```
lucaswerkmeister: I wonder if it would make sense to turn $1 into a list parameter?

> The only supported entity {{PLURAL:$2|type is|types are}} $1.

tests/phpunit/unit/InputToEntity/FullWikibaseItemInputTest.php
```diff
@@ -0,0 +1,31 @@
...
+use Deserializers\Deserializer;
+use MediaWiki\Extension\WikibaseReconcileEdit\InputToEntity\FullWikibaseItemInput;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\ReconciliationException;
+
+/**
+ * @covers \MediaWiki\Extension\WikibaseReconcileEdit\InputToEntity\FullWikibaseItemInput
+ */
+class FullWikibaseItemInputTest extends \MediaWikiUnitTestCase {
+
+	public function testItemType() {
```
lucaswerkmeister: I think it would be good to have a second test that calls `getItem()` with a different type (e.g. `[ 'type' => 'property' ]`) instead of an empty array. This could be done with a `@dataProvider`, since the rest of the test would be identical.

tests/phpunit/integration/MediaWiki/ExternalLinksTest.php
```diff
@@ -95,4 +96,20 @@ public function testPageIdsContainingNothing() {
 		$this->assertSame( [], $result );
 	}
 
+	public function testGetSingleIndex() {
+		$url = 'whatever';
+		$rows = $this->getExternalLinksRows( $url );
```
lucaswerkmeister: I think this and the next line can be removed. It doesn’t matter what’s in the database – `LinkFilter::makeindexes()` isn’t supposed to access the database – the important thing is that `$url` is protocol-relative (which is why `makeIndexes()` returns two elements: one for HTTP, one for HTTPS).

includes/MediaWiki/ExternalLinks.php
```diff
@@ -39,10 +40,12 @@ public function pageIdsContainingUrl( string $url ) : array {
 	 * @param string $url
 	 * @return string the el_index value
 	 */
-	private function getSingleIndexOrDie( string $url ) : string {
+	private function getSingleIndexOrThrowException( string $url ) : string {
 		$indexes = LinkFilter::makeIndexes( $url );
 		if ( count( $indexes ) !== 1 ) {
-			die( 'Unexpected issue with LinkFilter return' );
```
lucaswerkmeister, line 45: I think we don’t need to make this message translatable, actually. This isn’t really a case of a potential user error – if the returned array has more than one element, then the MediaWiki method has behaved differently than we expect. I think `Assert::invariant()` might be the most appropriate method here.

tests/phpunit/integration/MediaWiki/ExternalLinksTest.php
guergana: Done!.

includes/MediaWiki/ExternalLinks.php
guergana: Changed.

i18n/en.json
guergana: Done.

tests/phpunit/unit/InputToEntity/FullWikibaseItemInputTest.php
guergana: Good idea. Done.

i18n/qqq.json
```diff
@@ -16,5 +16,6 @@
...
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key"
+	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key",
+	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "Error returned by the API if the full wikibase item input's type parameter is not supported.\n\nParameters:\n $1 - the required types (comma-separated list)"
```
lucaswerkmeister: $2 – the number of supported types

(also, maybe change $1 from “required” to “supported” types?)

includes/InputToEntity/FullWikibaseItemInput.php
```diff
@@ -20,10 +22,13 @@ public function __construct( Deserializer $deserializer ) {
 	}
 
 	public function getItem( array $inputEntity ) : Item {
-		if ( !array_key_exists( 'type', $inputEntity ) || $inputEntity['type'] !== 'item' ) {
-			die( 'Only supported entity type is \'item\'' );
+		$supportedTypes = [ 'item' ];
+		if ( !array_key_exists( 'type', $inputEntity ) || !in_array( $inputEntity['type'], $supportedTypes ) ) {
+			throw new ReconciliationException(
+				MessageValue::new( 'wikibasereconcileedit-fullwikibaseiteminput-unsupported-type' )
+					->textListParams( $supportedTypes )
```
lucaswerkmeister, line 29: `->numParams( count( $supportedTypes ) )`

tests/phpunit/unit/InputToEntity/FullWikibaseItemInputTest.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class FullWikibaseItemInputTest extends \MediaWikiUnitTestCase {
+
+	public function provideTestGetItem() {
+		yield 'Empty' => [
+			[]
+		];
+		yield 'Type Property' => [
+			[
+				'wikibasereconcileedit-version' => '0.0.1/full',
```
lucaswerkmeister: The version isn’t used by the `FullWikibaseItemInput`, right? I’d leave it out to avoid confusion (it’s used by a different class, `EditRequestParser`, which we don’t test here).

includes/InputToEntity/FullWikibaseItemInput.php
guergana: Why is the number of supported types relevant??

lucaswerkmeister: It’s used in the message to decide whether it should say “type” or “types”.

i18n/qqq.json
guergana: Same as in the comment below

includes/InputToEntity/FullWikibaseItemInput.php
guergana: Oh, ok. :)

tests/phpunit/unit/InputToEntity/FullWikibaseItemInputTest.php
guergana: Yes, that's true.

lucaswerkmeister (approved):

PR #35, opened 2021-05-21
Add authentication

toban:


includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -34,16 +37,21 @@ class EditEndpoint extends SimpleHandler {
 	/** @var SimplePutStrategy */
 	private $simplePutStrategy;
 
+	/** @var User */
+	private $user;
```
lucaswerkmeister, line 36: Is this supposed to be a `User` or a `UserIdentity`? It’s `UserIdentity` in the constructor but `User` in the phpdoc comment.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -141,17 +157,15 @@ public function run() {
 				$otherItem->getItem(),
 				'Reconciliation Edit',
 				EDIT_NEW,
-				// TODO actually do a token check?
-				false
+				$this->user->getEditToken()
```
lucaswerkmeister: I don’t think this is the right way to do the token check – we’re getting the user’s token and then asking `EditEntity` to compare it to the user’s token, which should always succeed. The token should come from the request parameters, if I understand correctly.

lucaswerkmeister: Hm, in master of MediaWiki core, `UpdateHandler` and `EditHandler::getActionModuleToken()` have some interesting related code – apparently session providers can declare whether they’re potentially vulnerable to CSRF or not, and depending on that the `token` parameter is not required. (In fact, if the session declares itself safe but the request still includes a token, an error is raised, which strikes me as overly hostile to API consumers – but whatever.)

But that code is currently in `EditHandler`, a subclass of `ActionModuleBasedHandler`, so we probably couldn’t directly use it even if we were targeting MediaWiki master (rather than 1.35).

toban: Ok, have a look at https://github.com/wmde/WikibaseReconcileEdit/pull/35/commits/2f1132640bf84050be6050160d0143894462899b

It switches these parameters from being passed in as post parameters but rather all together in the body. Made the tests and the code to look a bit more similar to what is being done in UpdateHandler.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -197,19 +199,31 @@ public function needsWriteAccess() {
...
+			throw new HttpException( "Unsupported Content-Type",
+				415,
+				[ 'content_type' => $contentType ]
+			);
+		}
+
+		return new JsonBodyValidator( [
 			'entity' => [
-				self::PARAM_SOURCE => 'post',
+				self::PARAM_SOURCE => 'body',
```
lucaswerkmeister: This isn’t actually used by `JsonBodyValidator` as far as I can tell, and `'body'` is not one of the allowed values according to the constant documentation, so I think we should just remove these lines.

includes/MediaWiki/Request/EditRequest.php
```diff
@@ -17,23 +17,29 @@ class EditRequest {
...
 	public function __construct(
 		PropertyId $reconcilePropertyId,
 		Item $entity,
+		string $token,
 		array $otherItems = []
 	) {
 		$this->reconcilePropertyId = $reconcilePropertyId;
 		$this->entity = $entity;
 		$this->otherItems = $otherItems;
+		$this->token = $token;
```
lucaswerkmeister: (I’d prefer to move this up one line, so that the field, constructor params, assignments and getters are all in the same order.)

includes/MediaWiki/Request/EditRequestParser.php
```diff
@@ -39,11 +38,9 @@ public function __construct(
...
-		$reconcile = json_decode(
-			$request->getPostParams()['reconcile'],
-			true
-		);
+	public function parseRequestBody( array $requestBody ): EditRequest {
+		$reconcile = $requestBody['reconcile'];
+
 		if ( !is_array( $reconcile ) ) {
 			throw new LocalizedHttpException(
 				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-json' ),
```
lucaswerkmeister: This should probably be a different error message, since the JSON syntax is now checked by `JsonBodyValidator`.

includes/MediaWiki/Request/EditRequestParser.php
```diff
@@ -84,10 +81,7 @@ public function parseRequestInterface( RequestInterface $request ): EditRequest
...
 		}
 
-		$entity = json_decode(
-			$request->getPostParams()['entity'],
-			true
-		) ?: null;
+		$entity = $requestBody['entity'] ?: null;
 
 		if (
 			// TODO separate error message for invalid entity JSON?
```
lucaswerkmeister: This TODO comment is also obsolete in that form, I think.

includes/MediaWiki/Request/EditRequestParser.php
```diff
@@ -118,7 +112,8 @@ public function parseRequestInterface( RequestInterface $request ): EditRequest
 		return new EditRequest(
 			$reconcilePropertyId,
 			$inputEntity,
-			$otherItems
+			array_key_exists( 'token', $requestBody ) ? $requestBody['token'] : '',
```
lucaswerkmeister: This shouldn’t be necessary, I think, since we already specify a default value for the `JsonBodyValidator`.

tests/phpunit/unit/MediaWiki/Request/EditRequestParserTest.php
```diff
@@ -164,14 +164,14 @@ public function provideUnsupportedEntityVersion(): iterable {
 
 	/** @dataProvider provideInvalidReconcileJson */
 	public function testParseRequestInterface_invalidReconcileJson( string $reconcile ): void {
```
lucaswerkmeister: This test doesn’t make sense anymore – the request parser no longer decodes JSON.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -73,16 +72,25 @@ public static function factory(
 	}
 
 	public function run() {
-		if ( !$this->user->isRegistered() ) {
+		// Rest Validator returns null if submitted as form
```
lucaswerkmeister, line 72: Hm, are you sure? From looking at the code, it seems like `run()` should never even run if validation fails, at least when using MediaWiki’s `Router`.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -73,16 +72,25 @@ public static function factory(
...
+		$requestBody = $this->getValidatedBody();
+		if ( $requestBody === null ) {
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-json' ),
```
lucaswerkmeister: I don’t think this is the right message, since it’s not really the reconcile JSON, specifically, that’s invalid.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -203,9 +211,9 @@ public function needsWriteAccess() {
 	 */
 	public function getBodyValidator( $contentType ) {
 		if ( $contentType !== 'application/json' ) {
-			throw new HttpException( "Unsupported Content-Type",
-				415,
-				[ 'content_type' => $contentType ]
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-json' ),
```
lucaswerkmeister: This also doesn’t seem like the right use case for that message?

includes/MediaWiki/Api/EditEndpoint.php
toban: It exists before the body validator is run here https://github.com/wikimedia/mediawiki/blob/REL1_35/includes/Rest/Validator/Validator.php#L163

lucaswerkmeister: :roll_eyes: I see, thanks.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -57,14 +65,32 @@ public static function factory(
...
+		// Rest Validator returns null if submitted as form
+		$requestBody = $this->getValidatedBody();
+		if ( $requestBody === null ) {
+			throw new LocalizedHttpException(
+				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-reconcile-json' ),
+				415
+			);
+		}
+
+		if ( !$this->user->isRegistered() || !$this->user->matchEditToken( $requestBody['token'] ) ) {
```
lucaswerkmeister: If we’re checking the edit token here anyways (which I think is a good idea, because it happens before we start handing out new item IDs), then we could probably continue to pass `false` into `attemptSave()`? Or is it useful to do the token check twice?

includes/MediaWiki/Api/EditEndpoint.php
toban: done

includes/MediaWiki/Api/EditEndpoint.php
toban: done

includes/MediaWiki/Api/EditEndpoint.php
toban: done, added a test this validation too.

includes/MediaWiki/Request/EditRequest.php
toban: done

tests/phpunit/unit/MediaWiki/Request/EditRequestParserTest.php
toban: done

includes/MediaWiki/Request/EditRequestParser.php
toban: done

includes/MediaWiki/Api/EditEndpoint.php
toban: Feels a bit backwards not using the token when saving if it was included for this purpose?

lucaswerkmeister: Fair enough ^^

i18n/en.json
```diff
@@ -15,7 +15,10 @@
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-editendpoint-invalid-request-body": "Invalid or empty JSON request body",
+	"wikibasereconcileedit-editendpoint-invalid-content-type": "Content-type should be application/json",
```
lucaswerkmeister: Hm, maybe the expected content type should be a message parameter?

i18n/en.json
```diff
@@ -15,7 +15,10 @@
...
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
 	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
-	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1."
+	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
+	"wikibasereconcileedit-unauthorized-access": "Not logged in."
```
lucaswerkmeister: Is it intentional that this doesn’t mention the possibility of an invalid CSRF token?

README.md
```diff
@@ -104,6 +104,19 @@ Edit payload `0.0.1/full` should look like this:
...
 **Note: that you do not need to provide statement GUIDs or any hashes.**
 
+## Authentication
+
+In order to make requests they need to be authenticated and with an [edit token](https://www.mediawiki.org/wiki/Manual:Edit_token) supplied in the request payload.
+Authentication can either be done by using the [action api](https://www.mediawiki.org/wiki/API:Login) or OAuth.
+
+OAuth requires the [OAuth Extension](https://www.mediawiki.org/wiki/Extension:OAuth) to be installed and a OAuth consumer created via the `wiki/Special:OAuthConsumerRegistration/propose` page.
+
+Follow the instructions [here]((https://www.mediawiki.org/wiki/OAuth/Owner-only_consumers)) on how to create a OAuth owner-only consumer.
```
lucaswerkmeister: There’s an extra set of `()` here, which apparently causes GitHub to not render this as a link at all?

README.md
```diff
@@ -104,6 +104,19 @@ Edit payload `0.0.1/full` should look like this:
...
+
+For an example on how to request a edit token and make an authenticated request using OAuth see the [example python script](example/oauth/main.py).
+
+**Note: OAuth owner-only consumers has not been confirmed to work without https**
```
lucaswerkmeister: Is this still the case?

toban: I wasn't able to get it working :|

README.md
toban: done

i18n/en.json
toban: sure!

i18n/en.json
toban: I updated to be a bit more specific what could be wrong.

tests/phpunit/unit/MediaWiki/Request/EditRequestParserTest.php
```diff
@@ -122,14 +123,14 @@ public function testParseRequestInterface_good(): void {
 	}
```
lucaswerkmeister, line 122: There should probably be an assertion here about the `$editRequest->token()`? (Unless you want to squash-merge the PR, in which case it’s moot because the `token()` goes away again anyways ^^)

i18n/qqq.json
```diff
@@ -15,7 +15,11 @@
 	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Error returned by the API if the “entity” request parameter specifies qualifiers and/or references. They are not yet supported.",
 	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Error returned by the API if the “entity” request parameter doesn't contain a statement that uses the reconciliation property",
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
+	"wikibasereconcileedit-editendpoint-invalid-request-body": "Error returned by the API if the JSON request body is invalid or empty.",
+	"wikibasereconcileedit-editendpoint-invalid-content-type": "Error returned by the API if the request content type is not application/json",
```
lucaswerkmeister: This should probably explain what `$1` is and not mention `application/json`.

tests/phpunit/integration/MediaWiki/Api/EditEndpointTest.php
```diff
@@ -302,4 +304,10 @@ private function countItemsInDatabase(): int {
 		);
 	}
 
+	private function getMockSessionProvider( bool $safeAgainstCSRF ): SessionProviderInterface {
```
lucaswerkmeister, line 369: Hm, there’s no test that passes `false` here?

i18n/qqq.json
lucaswerkmeister: Done.

tests/phpunit/unit/MediaWiki/Request/EditRequestParserTest.php
lucaswerkmeister: Done.

tests/phpunit/integration/MediaWiki/Api/EditEndpointTest.php
lucaswerkmeister: Added a test, I hope it looks okay.

README.md
lucaswerkmeister: I tested it locally and it worked fine for me.

lucaswerkmeister (approved):

tests/phpunit/integration/MediaWiki/Api/EditEndpointTest.php
toban: nice :+1:

toban:
LGTM!

PR #36, opened 2021-05-25
Lookup properties through label

toban:


includes/ServiceWiring.php
```diff
@@ -53,7 +58,52 @@
 		return new MinimalItemInput(
 			$repo->getPropertyDataTypeLookup(),
 			$repo->getValueParserFactory(),
-			WikibaseReconcileEditServices::getReconciliationService( $services )
+			WikibaseReconcileEditServices::getReconciliationService( $services ),
+			WikibaseReconcileEditServices::getPropertyLabelResolver( $services ),
+			LoggerFactory::getInstance( 'WikibaseReconcileEdit.MinimalItemInput' )
```
lucaswerkmeister: This seems too specific to me – each log channel has to be configured in the site config, after all. `'WikibaseReconcileEdit'` should be enough, I think.

includes/InputToEntity/MinimalItemInput.php
```diff
@@ -113,6 +130,45 @@ public function getItem( array $inputEntity, PropertyId $reconcileUrlProperty )
...
+	 * @param string $propertyId
+	 * @return PropertyId|null
+	 * @throws PropertyLabelNotResolvedException
+	 */
+	public function getPropertyId( string $propertyId ): ?PropertyId {
+		if ( strlen( $propertyId ) > 1 && $propertyId[0] === 'P' ) {
+			try {
+				return new PropertyId( $propertyId );
+			} catch ( \InvalidArgumentException $exception ) {
+				$this->logger->warning( 'Tried parsing something weird' );
```
lucaswerkmeister: I don’t think this warrants a warning – it’s totally reasonable for a property label to start with `P` (Wikidata has [328 such properties](https://query.wikidata.org/#SELECT%20%2a%20WHERE%20%7B%0A%20%20%3Fp%20a%20wikibase%3AProperty%3B%0A%20%20%20%20%20rdfs%3Alabel%20%3FpLabel.%0A%20%20FILTER%28LANG%28%3FpLabel%29%20%3D%20%22en%22%29%0A%20%20FILTER%28STRSTARTS%28%3FpLabel%2C%20%22P%22%29%29%0A%7D) just in English). I think we should try to parse any string as a property ID first, and if that fails, then fall back to interpreting the string as a label, with at most an `->info()` log message (or maybe even just `->debug()`).

lucaswerkmeister (commented):
Some early comments. Overall, this looks pretty good to me, though the copied `PropertyLabelResolver` service wiring is perhaps unsatisfying.

includes/ServiceWiring.php
guergana: Updated!

includes/InputToEntity/MinimalItemInput.php
guergana: And after interpreting it as a label do we just use the label in the info or debug message?

guergana: This has been addressed. Tho I am not sure if what I did is what you meant. Could you check again?

includes/InputToEntity/MinimalItemInput.php
```diff
@@ -113,6 +130,43 @@ public function getItem( array $inputEntity, PropertyId $reconcileUrlProperty )
...
+	/**
+	 * @param string $label
+	 * @return PropertyId|null
+	 * @throws PropertyLabelNotResolvedException
+	 */
+	public function getPropertyIdByLabel( string $label ): ?PropertyId {
+		$entityIds = $this->propertyLabelResolver->getPropertyIdsForLabels( [ $label ] );
+
+		if ( empty( $entityIds ) ) {
+			throw new PropertyLabelNotResolvedException( $label, 'en' );
```
lucaswerkmeister: That’s a WikibaseClient exception, and it also uses a message key that only exists in client. If we want to depend on Repo only, I think we should throw a `ReconciliationException` here, with a message of our own.

includes/InputToEntity/MinimalItemInput.php
lucaswerkmeister: Yup, that looks good now. Though I feel like we could even remove the logger call completely, and then stop injecting a logger and save ourselves a bunch of code. WDYT?

guergana: :shrug: ok!

lucaswerkmeister:
Open TODOs I remember:

- add tests in `MinimalItemInputTest`
- look into the `mockPropertyLabelResolver()` in `MinimalItemInputTest` and `EditRequestParserTest`, those look weird to me

includes/InputToEntity/MinimalItemInput.php
guergana: This seems resolved.

includes/ServiceWiring.php
```diff
@@ -9,7 +9,11 @@
 use MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ItemReconciler;
 use MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ReconciliationService;
 use MediaWiki\MediaWikiServices;
+use Wikibase\DataModel\Entity\Property;
 use Wikibase\DataModel\Services\Statement\GuidGenerator;
+use Wikibase\Lib\Store\Sql\Terms\CachedDatabasePropertyLabelResolver;
```
guergana, line 14: @toban and I worked on this and we have some doubts/comments: The `CachedDatabasePropertyLabelResolver` is not yet working with Federated Properties. We tested it. We could build our own or make it work with wikibase. `CachedDatabasePropertyLabelResolver` is only used in client. What approach do you think we should take @lucaswerkmeister

lucaswerkmeister: Let’s merge this without FP support and then see if we need it later.

i18n/qqq.json
```diff
@@ -18,5 +18,6 @@
 	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Error returned by the API if the “entity” request parameter if the reconciliation statement's mainsnak is not of type value",
 	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "Error returned by the API if the reconciliation request matches multiple items.\n\nParameters:\n* $1 – the matched items (comma-separated list)\n* $2 – the number of matched items",
 	"wikibasereconcileedit-minimaliteminput-required-keys": "Error returned by the API if the minimal item input statements parameter is missing a certain key.\n\nParameters:\n $1 - the missing key",
+	"wikibasereconcileedit-minimaliteminput-found-multiple-properties" : "Error returned by the API if the minimal item input result found multiple matching properties with the same label.\n\nParameters:\n $1 - the property ids",
```
lucaswerkmeister: This message needs another parameter with the number of properties, since some languages have different plural forms depending on the number.

includes/InputToEntity/MinimalItemInput.php
```diff
@@ -148,12 +148,13 @@ public function getPropertyIdByLabel( string $label ): PropertyId {
 		}
 
 		if ( count( $entityIds ) !== 1 ) {
-			throw new \InvalidArgumentException( "Found multiple properties" );
+			throw new ReconciliationException(
+				MessageValue::new( 'wikibasereconcileedit-minimaliteminput-found-multiple-properties' )
```
lucaswerkmeister: Hm, didn’t we say in another call that this didn’t need a translatable message since it should never happen?

I think we should remove this check completely, in fact. I think there should just be one condition: `isset( $entityIds[ $label ] )` – if yes, then return it, otherwise “property not found”. We don’t need to care if there are any other entries in the array. (On the other hand, currently we don’t actually check that `$entityIds[ $label ]` is set – we just assume that if the array has one element, it’ll be that one.)

i18n/qqq.json
lucaswerkmeister: (If we keep the message at all, that is. See below.)

includes/ServiceWiring.php
```diff
@@ -53,7 +57,52 @@
...
+			WikibaseReconcileEditServices::getPropertyLabelResolver( $services ),
+		);
+	},
+
+	'WikibaseReconcileEdit.PropertyLabelResolver' => function ( MediaWikiServices $services ):
+		CachedDatabasePropertyLabelResolver {
+		$repo = WikibaseRepo::getDefaultInstance();
+		$settings = $repo->getSettings();
+
+		$languageCode = $services->getContentLanguage()->getCode();
```
toban, line 70: Maybe this should be a setting for the WikibaseReconcileEdit extension rather than defaulting to the language of the mediawiki installation?

tests/api-testing/REST/EditEndpoint.js
```diff
@@ -103,4 +104,58 @@ describe( 'POST /edit', () => {
...
+		];
+
+		const params = requestHelper.getRequestPayload(
+			requestStatements,
+			reconciliationPropertyId
+		);
+		const { status, body } = await client.post(
+			'/edit',
+			params,
+			{ 'content-type': 'application/x-www-form-urlencoded' } // why?
```
toban: this would crash in CI now, needs a rebase

includes/ServiceWiring.php
toban: lets not

includes/InputToEntity/MinimalItemInput.php
toban: all good points. We were talking about this but couldn't point to where this uniqueness guarantee(property label, language) comes from in wikibase

tests/api-testing/REST/EditEndpoint.js
toban: done

i18n/qqq.json
toban: removed

includes/InputToEntity/MinimalItemInput.php
toban: removed

toban:
LGTM!

includes/ServiceWiring.php
guergana: Sounds ok to me :)

lucaswerkmeister (approved):
Added a minor improvement, looks good to me otherwise. Should be good to squash-merge :+1:

PR #37, opened 2021-05-25
Extract ItemReconciler service

lucaswerkmeister:
This service wraps a `ReconciliationService` and `SimplePutStrategy` and extracts the reconciliation URL from an input item before passing it to the `ReconciliationService` and `SimplePutStrategy`. Its return value is an Item and a revision ID, but the meaning is quite different, so we introduce a new result class, `ReconciledItem`, rather than reusing `ReconciliationItem`.

Bug: [T283567](https://phabricator.wikimedia.org/T283567)

---

Includes one preparation commit to move some classes.

The names “ItemReconciler” and “ReconciledItem” aren’t that great… any other ideas?


includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -153,14 +111,14 @@ public function run() {
...
-			$reconciliationItem->getRevision()
+			$reconciledItem->getBaseRevisionId()
 		);
 
 		if ( $saveStatus->isOK() ) {
 			$saveStatus->merge( $editEntity->attemptSave(
 				$toSave,
 				'Reconciliation Edit',
-				$reconciliationItem->getRevision() ? null : EDIT_NEW,
+				$reconciledItem->isNew() ? EDIT_NEW : 0,
```
toban: Shouldn't this be `EDIT_UPDATE`?

Seems like the 0 or null becomes a `EDIT_AUTOSUMMARY` by default? Not sure if it even matters.

includes/Reconciliation/ReconciliationItem.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This class contains an item as it was found in the database,
+ * or a newly created item if no existing item was found,
+ * in which case the revision ID is false.
+ * It should not be confused with {@link ReconciledItem},
+ * which includes modifications made to the item based on the input.
+ *
+ * @license GPL-2.0-or-later
+ */
+class ReconciliationItem {
```
toban: How do you feel about naming this `ReconciliationServiceItem` ?

Even though we have the comment not to confuse the two i think making the difference a bit more obvious will make this less likely to be confused?

toban (approved):
Looks good to me, just a few thoughts that popped up.

includes/MediaWiki/Api/EditEndpoint.php
lucaswerkmeister: Hm, `EDIT_UPDATE` might make sense. But where do you see `EDIT_AUTOSUMMARY` coming in?

includes/Reconciliation/ReconciliationItem.php
lucaswerkmeister: Sounds okay to me. Or maybe even `ReconciliationServiceResult`, without “item”?

includes/MediaWiki/Api/EditEndpoint.php
toban: I thought it was set here https://github.com/wikimedia/Wikibase/blob/REL1_35/repo/includes/EditEntity/MediawikiEditEntity.php#L739 if left out

lucaswerkmeister: Okay, but it doesn’t look like we can influence this, so I’d leave it out of this call.

includes/Reconciliation/ReconciliationItem.php
lucaswerkmeister: Okay, I’ll insert another commit to rename it to `ReconciliationServiceResult`.

lucaswerkmeister: Hm, on the other hand the term “item” is used a lot for these results (e.g. `EditRequest::otherItems()`), and changing those to “result” would probably make it less clear overall.

Let’s go with `ReconciliationServiceItem`.

PR #38, opened 2021-05-25
Fix WikibaseReconcileEdit saves statements without a statement ID

guergana:
Add guid when creating new statements

Bug: T283358

### DIFF
--- a/includes/Reconciliation/ReconciliationService.php
+++ b/includes/Reconciliation/ReconciliationService.php
@@ -9,6 +9,7 @@ use TitleFactory;
 use Wikibase\DataModel\Entity\Item;
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Services\Statement\GuidGenerator;
 use Wikibase\DataModel\Snak\PropertyValueSnak;
 use Wikibase\Lib\Store\EntityIdLookup;
 use Wikibase\Lib\Store\EntityRevisionLookup;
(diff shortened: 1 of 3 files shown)


includes/MediaWiki/ReconciliationService.php
```diff
@@ -146,9 +147,15 @@ public function getOrCreateItemByStatementUrl(
 			$base->setId( ItemId::newFromNumber( $this->idGenerator->getNewId( 'wikibase-item' ) ) );
 			$reconciliationItem = new ReconciliationItem( $base, false );
 
+			$guidGenerator = new GuidGenerator();
```
lucaswerkmeister: The GUID generator should be injected as a dependency, in my opinion. (We already inject it into `SimplePutStrategy`.)

lucaswerkmeister: Also, if the GUID generator is injected, we can use a mocked generator in the tests which always returns `"$id\$00000000-0000-0000-0000-000000000000"`, or something like that, to avoid including part of the actual API response in the expected / comparison value.

guergana: Done.

guergana: > Also, if the GUID generator is injected, we can use a mocked generator in the tests which always returns `"$id\$00000000-0000-0000-0000-000000000000"`, or something like that, to avoid including part of the actual API response in the expected / comparison value.

Wait, this message wasn't here when I was adding the 'Done' comment

guergana: Now done I think. Please re-check :)

lucaswerkmeister (approved):

lucaswerkmeister:
Rebased due to merge conflicts.

PR #39, opened 2021-05-28
Update readme with new API updates

toban:
- Item statement reconciliation
- Property lookup by label

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This code is currently **work in progress** and some things are thus **known to 
 * 0.0.1/minimal statement value parsing options (such as language), could lead to "interesting" things
 * Anything to do with references, qualifiers
 * Statement ranks and sitelink badges may have an undetermined behavior
-* Editing always has a fixed user and summary currently, and token validation etc does not happen
+* The property lookup by name feature is not working yet with Federated Properties. [See below](#property-lookup-by-name).
 * Not fully tested, so there could be other bugs...
 
 ## /wikibase-reconcile-edit/v0/edit (Editing API)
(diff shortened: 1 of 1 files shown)


guergana:
It looks great :heart_eyes: Tho, as per [the ticket](https://phabricator.wikimedia.org/T283926) I would add a mention to Federated properties to the list named "This code is currently work in progress and some things are thus known to not to work:"

README.md
```diff
@@ -115,6 +115,62 @@ Follow the instructions [here](https://www.mediawiki.org/wiki/OAuth/Owner-only_c
 
 For an example on how to request a edit token and make an authenticated request using OAuth see the [example python script](example/oauth/main.py).
 
+## Property lookup by name
```
lucaswerkmeister: I think both “property lookup by name” and “statement reconciliation” should be subsections of the minimal input format, since we don’t support them for the full input format.

lucaswerkmeister (commented):
> It looks great heart_eyes Tho, as per [the ticket](https://phabricator.wikimedia.org/T283926) I would add a mention to Federated properties to the list named "This code is currently work in progress and some things are thus known to not to work:"

Also, several other items of that list should be removed by now.

README.md
guergana: Good point.

README.md
```diff
@@ -9,7 +9,8 @@ This code is currently **work in progress** and some things are thus **known to
 * 0.0.1/minimal statement value parsing options (such as language), could lead to "interesting" things
 * Anything to do with references, qualifiers
 * Statement ranks and sitelink badges may have an undetermined behavior
-* Editing always has a fixed user and summary currently, and token validation etc does not happen
+* Property lookup by name and Statement reconciliation are currently only supported for the minimal input format.
```
lucaswerkmeister: Hm, I don’t think that’s a limitation? It wouldn’t make sense to support them in the full input format IMHO, that format is supposed to be identical to the Wikibase JSON.

guergana: :sweat_smile: ok.

toban:
Nice, reads and looks good to me :+1:

PR #41, opened 2021-05-31
Batch editing

toban:


toban:
- [x] savedOtherItems might reference not needed
- [x] No exception add break
- [x] Add result only if status is good or ok
- [ ] change user to private on EditEndpoint
- [ ] persistItem can set the item revision instead of returning saved items.
- [x] just return an array on parseBatchRequests - remove BatchEditRequest
- [x] Add PHP integration test for batch-edit endpoint

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -52,69 +53,31 @@ public function __construct(
...
-			$reconciledItem = $this->itemReconciler->reconcileItem(
-				$inputEntity,
-				$reconcileUrlProperty
-			);
-		} catch ( ReconciliationException $rex ) {
-			throw new LocalizedHttpException( $rex->getMessageValue(), 400 );
-		}
-
-		// And make the edit
+	public function persistItem( ReconciledItem $reconciledItem, array $otherItems ): array {
```
toban, line 83: Add PHPDOC

lucaswerkmeister: Done

lucaswerkmeister:
> just return an array on parseBatchRequests - remove BatchEditRequest

done, but all of the tests still passed even though I hadn’t updated them, which means

- [x] add tests for parseBatchRequests

lucaswerkmeister:
on the other points:

> change user to private on EditEndpoint

not convinced this one is needed tbh. I’d be willing to merge without this

> persistItem can set the item revision instead of returning saved items.

to be done in a follow-up

lucaswerkmeister (approved):

guergana:
LGTM. Also nice now we new tests.

includes/MediaWiki/Api/BatchEditEndpoint.php
```diff
@@ -0,0 +1,125 @@
...
+			$response['results'] = array_map( static function ( $result ) {
+				/** @var EntityRevision $entityRevision */
+				$entityRevision = $result['revision'];
+				return [
+					'entityId' => $entityRevision->getEntity()->getId()->getSerialization(),
+					'revisionId' => $entityRevision->getRevisionId(),
+				];
+			}, $status->getValue() );
+		}
+		// TODO attach errors
```
toban, line 101: We could do this now maybe or write this down as a ticket? I suspect there might be a series of followups planned already?

tests/phpunit/integration/MediaWiki/Api/BatchEditEndpointTest.php
```diff
@@ -0,0 +1,263 @@
...
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Snak\PropertyValueSnak;
+use Wikibase\Repo\WikibaseRepo;
+
+/**
+ * @covers MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Api\BatchEditEndpoint
+ * @group Database
+ * @license GPL-2.0-or-later
+ */
+class BatchEditEndpointTest extends \MediaWikiIntegrationTestCase {
```
toban, line 28: This reminds me a lot of EditEndpoint test, not sure there are any real benefits of doing so but maybe we could refactor these afterwards to share some of this code / token tests etc if we feel there is enough time.

tests/phpunit/unit/MediaWiki/Request/EditRequestParserTest.php
```diff
@@ -97,7 +97,7 @@ private function getPropertyDataTypeLookup(): PropertyDataTypeLookup {
 		return $propertyDataTypeLookup;
 	}
 
-	public function testParseRequestInterface_good(): void {
+	public function testParseRequestBody_good(): void {
```
toban, line 100: good catch

toban (commented):
Can't approve this but I do! it looks great. Just some comments for possible followups.

includes/MediaWiki/Api/BatchEditEndpoint.php
lucaswerkmeister: Ticket makes sense, I think. (The same comment applies to `SingleEditEndpoint`, though we don’t have a TODO there.)

toban: https://phabricator.wikimedia.org/T284134

PR #45, opened 2021-06-02
Rename i18n messages

guergana:
Bug: T284051


lucaswerkmeister (commented):
I don’t think we should put the class name into all of those messages – that just means the message key will become outdated again the next time we move the code around. I think the message keys should try to be more general. WDYT?

guergana:
> I don’t think we should put the class name into all of those messages – that just means the message key will become outdated again the next time we move the code around. I think the message keys should try to be more general. WDYT?

It's a good idea. Tho, to be fair it also makes the messages much clearer than without the class name... Changing the name of the class is also not a big deal. :thinking:

guergana:
> I don’t think we should put the class name into all of those messages – that just means the message key will become outdated again the next time we move the code around. I think the message keys should try to be more general. WDYT?

It's done!

i18n/en.json
```diff
@@ -6,20 +6,20 @@
...
-	"wikibasereconcileedit-editendpoint-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
-	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
-	"wikibasereconcileedit-editendpoint-invalid-request-body": "Invalid or empty JSON request body",
-	"wikibasereconcileedit-editendpoint-invalid-content-type": "Content-type should be $1",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
-	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
+	"wikibasereconcileedit-invalid-content-type": "Content-type should be $1",
+	"wikibasereconcileedit-invalid-request-body": "Invalid or empty JSON request body",
```
lucaswerkmeister: If this mentions JSON in the message text, maybe it should be invalid-_json_-request-body?

i18n/en.json
```diff
@@ -6,20 +6,20 @@
...
+	"wikibasereconcileedit-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
+	"wikibasereconcileedit-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
+	"wikibasereconcileedit-invalid-reconcile-parameter": "Invalid reconcile parameter. Expecting an array",
+	"wikibasereconcileedit-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-unspecified-entity-input-version": "Entity input version must be specified",
+	"wikibasereconcileedit-unsupported-reconcile-version": "The only supported reconciliation {{PLURAL:$2|version is|versions are}} $1",
+	"wikibasereconcileedit-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
```
lucaswerkmeister: “unsupported type” in the message key sounds overly generic to me – how about unsupported-_entity_-type?

i18n/en.json
```diff
@@ -6,20 +6,20 @@
...
+	"wikibasereconcileedit-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
+	"wikibasereconcileedit-qualifiers-references-not-supported": "Qualifiers and References are not currently supported",
+	"wikibasereconcileedit-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
+	"wikibasereconcileedit-property-not-found": "Could not find property with label '$1'",
+	"wikibasereconcileedit-required-keys": "Statements missing '$1' key",
```
lucaswerkmeister: “required keys” also sounds very generic. Maybe statement-required-key or statement-missing-key? (With “key” instead of “keys” because it complains about a single key AFAICT.)

lucaswerkmeister (commented):
Some comments on individual messages that I think we can improve at the same time.

i18n/en.json
guergana: All good points :)

i18n/en.json
guergana: addressed

i18n/en.json
guergana: addressed :)

i18n/en.json
```diff
@@ -5,22 +5,21 @@
...
-	"wikibasereconcileedit-editendpoint-reconciliation-property-missing-in-statements": "Entity must have at least one statement for the reconciliation Property",
-	"wikibasereconcileedit-editendpoint-invalid-reconciliation-statement-type": "Reconciliation statement must contain a value",
-	"wikibasereconcileedit-editendpoint-invalid-request-body": "Invalid or empty JSON request body",
-	"wikibasereconcileedit-editendpoint-invalid-content-type": "Content-type should be $1",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
-	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
-	"wikibasereconcileedit-unauthorized-access": "Unauthorized access. You need to be registered and provide a valid token.",
-	"wikibasereconcileedit-property-datatype-lookup-error": "Could not find datatype for property with id '$1'"
+	"wikibasereconcileedit-desc": "This is the ON OR Prototype extension",
```
lucaswerkmeister: This seems to revert #51?

i18n/en.json
```diff
@@ -5,22 +5,21 @@
...
+	"wikibasereconcileedit-invalid-content-type": "Content-type should be $1",
+	"wikibasereconcileedit-invalid-JSON-request-body": "Invalid or empty JSON request body",
+	"wikibasereconcileedit-invalid-entity-input-version": "Unknown entity input version $1; the only supported {{PLURAL:$3|version is|versions are}} $2",
+	"wikibasereconcileedit-invalid-property-type-must-be-url": "$1 property $2 must be of type url. The datatype of the provided property is $3",
+	"wikibasereconcileedit-invalid-reconcile-parameter": "Invalid reconcile parameter. Expecting an array",
+	"wikibasereconcileedit-invalid-reconcile-propertyid": "$1 requires a single $2 key mapped to a property ID, such as P123",
+	"wikibasereconcileedit-unspecified-entity-input-version": "Entity input version must be specified",
+	"wikibasereconcileedit-unsupported-reconcile-version": "The only supported reconciliation {{PLURAL:$2|version is|versions are}} $1",
+	"wikibasereconcileedit-unsupported-entity-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
```
lucaswerkmeister, line 17: I guess this should move one line up now, if you want to keep the alphabetic order?

i18n/en.json
```diff
@@ -5,22 +5,21 @@
...
-	"wikibasereconcileedit-editendpoint-invalid-request-body": "Invalid or empty JSON request body",
-	"wikibasereconcileedit-editendpoint-invalid-content-type": "Content-type should be $1",
-	"wikibasereconcileedit-reconciliationservice-matched-multiple-items": "The reconciliation is ambiguous; matched {{PLURAL:$2|item|$2 items}}: $1",
-	"wikibasereconcileedit-minimaliteminput-required-keys": "Statements missing '$1' key",
-	"wikibasereconcileedit-fullwikibaseiteminput-unsupported-type": "The only supported entity {{PLURAL:$2|type is|types are}} $1.",
-	"wikibasereconcileedit-unauthorized-access": "Unauthorized access. You need to be registered and provide a valid token.",
-	"wikibasereconcileedit-property-datatype-lookup-error": "Could not find datatype for property with id '$1'"
+	"wikibasereconcileedit-desc": "This is the ON OR Prototype extension",
+	"wikibasereconcileedit-invalid-content-type": "Content-type should be $1",
+	"wikibasereconcileedit-invalid-JSON-request-body": "Invalid or empty JSON request body",
```
lucaswerkmeister: Likewise, maybe move this down? (Also, I’d prefer lowercase “json” in the message key, AFAIK uppercase characters in message keys are pretty unusual.)

i18n/en.json
guergana: no alphabetical order

i18n/en.json
guergana: yes, indeed, didn't realize there was a different PR to update messages.

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -60,7 +60,7 @@ public function run() {
 
 		if ( $this->requestBody === null ) {
 			throw new LocalizedHttpException(
-				MessageValue::new( 'wikibasereconcileedit-editendpoint-invalid-request-body' ),
+				MessageValue::new( 'wikibasereconcileedit-invalid-JSON-request-body' ),
```
lucaswerkmeister: this JSON needs to be lowercase now

i18n/en.json
lucaswerkmeister: okay, then it’s a bit unfortunate IMHO that the patch still reorders so many messages :/

includes/MediaWiki/Api/EditEndpoint.php
guergana: :scream: omg true.. this is what you get from trying to fix things during meetings :D i hope it's ok now

lucaswerkmeister (approved):

PR #46, opened 2021-06-02
Set revision information on ReconciliationServiceItem after saving it

lucaswerkmeister:
Bug: [T284068](https://phabricator.wikimedia.org/T284068)

### DIFF
--- a/includes/MediaWiki/Api/BatchEditEndpoint.php
+++ b/includes/MediaWiki/Api/BatchEditEndpoint.php
@@ -4,7 +4,6 @@ namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Api;
 
 use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequestParser;
 use MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ItemReconciler;
-use MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ReconciliationServiceItem;
 use MediaWiki\Extension\WikibaseReconcileEdit\ReconciliationException;
 use MediaWiki\Rest\LocalizedHttpException;
 use MediaWiki\Rest\Validator\JsonBodyValidator;
@@ -61,23 +60,14 @@ class BatchEditEndpoint extends EditEndpoint {
 		}
 
 		$status = Status::newGood( [] );
-		$savedOtherItems = [];
 		foreach ( $requests as $request ) {
 			$reconciledItem = $this->itemReconciler->reconcileItem(
 				$request->entity(),
 				$request->reconcilePropertyId()
 			);
+			$otherItems = $request->otherItems();
 
-			// only save new items once
-			$otherItems = array_filter(
-				$request->otherItems(),
-				function ( ReconciliationServiceItem $item ) use( $savedOtherItems ) {
-					return !in_array( $item->getItem()->getId(), $savedOtherItems );
-				}
-			);
-
-			[ $otherStatus, $itemIds ] = $this->persistItem( $reconciledItem, $otherItems );
-			$savedOtherItems = array_merge( $savedOtherItems, $itemIds );
+			$otherStatus = $this->persistItem( $reconciledItem, $otherItems );
 			if ( !$otherStatus->isOk() ) {
 				break;
 			}
--- a/includes/Reconciliation/ReconciliationServiceItem.php
+++ b/includes/Reconciliation/ReconciliationServiceItem.php
@@ -3,6 +3,8 @@
 namespace MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation;
 
 use Wikibase\DataModel\Entity\Item;
+use Wikibase\Lib\Store\EntityRevision;
+use Wikimedia\Assert\Assert;
 
 /**
  * An item found by the {@link ReconciliationService}, together with its revision ID.
@@ -43,6 +45,23 @@ class ReconciliationServiceItem {
 	 */
 	public function getRevision() {
 		return $this->revision;
+	}
+
+	/**
+	 * Replace the item and revision ID with those of the given entity revision.
+	 *
+	 * This should be used after the caller has saved the reconciliation service item.
+	 * The entity revision is assumed to be based on the previous {@link #getItem item};
+	 * in particular, the caller must ensure that the entity is an Item,
+	 * and that it has the same URL statement
+	 * (i.e. it is correct for a {@link ReconciliationService} to still return
+	 * this {@link ReconciliationServiceItem} instance for the same URL).
+	 */
+	public function updateFromEntityRevision( EntityRevision $entityRevision ): void {
+		$item = $entityRevision->getEntity();
+		Assert::parameterType( Item::class, $item, '$entityRevision->getEntity()' );
+		$this->item = $item;
+		$this->revision = $entityRevision->getRevisionId();
 	}
 
 }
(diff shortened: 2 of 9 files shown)


lucaswerkmeister:
> Based on #42 because otherwise I wouldn’t be able to run the API integration tests locally. I guess I could also rebase this to remove that commit if anyone would prefer that.

Nevermind, #42 was merged in the meantime :D

lucaswerkmeister:
I also tested this:

```diff
diff --git a/tests/phpunit/integration/MediaWiki/Api/SingleEditEndpointTest.php b/tests/phpunit/integration/MediaWiki/Api/SingleEditEndpointTest.php
index ccb3dce..8723d4b 100644
--- a/tests/phpunit/integration/MediaWiki/Api/SingleEditEndpointTest.php
+++ b/tests/phpunit/integration/MediaWiki/Api/SingleEditEndpointTest.php
@@ -287,10 +287,6 @@ public function testItemStatementReconciliation( array $requests, array $expecte
 			$item = WikibaseRepo::getDefaultInstance()->getEntityLookup()->getEntity( $itemId );
 			$this->assertInstanceOf( Item::class, $item );
 
-			// this needs to be reset in-order for the per-request "cache" to clear
-			// @see ReconciliationService::$items
-			MediaWikiServices::getInstance()->resetServiceForTesting( 'WikibaseReconcileEdit.ReconciliationService' );
-
 			$expectedRequestItems = $expectedItems[$requestIndex];
 
 			$this->assertEquals( $this->countItemsInDatabase(), count( $expectedRequestItems ) );
```

The test still passes, which means that the reconciliation service now behaves correctly when used for multiple requests in a row. (Though we still want to clear the cache for more realistic test behavior.)

includes/MediaWiki/Api/BatchEditEndpoint.php
```diff
@@ -61,23 +60,16 @@ public function run() {
...
 
 		$status = Status::newGood( [] );
-		$savedOtherItems = [];
 		foreach ( $requests as $request ) {
 			$reconciledItem = $this->itemReconciler->reconcileItem(
 				$request->entity(),
 				$request->reconcilePropertyId()
 			);
 
 			// only save new items once
```
toban: I don't think this make much sense here now?

includes/Reconciliation/ReconciliationServiceItem.php
```diff
@@ -45,4 +47,21 @@ public function getRevision() {
...
+	 * Replace the item and revision ID with those of the given entity revision.
+	 *
+	 * This should be used after the caller has saved the reconciliation service item.
+	 * The entity revision is assumed to be based on the previous {@link #getItem item};
+	 * in particular, the caller must ensure that the entity is an Item,
+	 * and that it has the same URL statement
+	 * (i.e. it is correct for a {@link ReconciliationService} to still return
+	 * this {@link ReconciliationServiceItem} instance for the same URL).
+	 */
+	public function replaceWith( EntityRevision $entityRevision ) {
```
toban: What do you think about naming this updateFromRevision or something similar? I find replaceWith a bit confusing. Not a big deal though.

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
```diff
@@ -0,0 +1,92 @@
...
+		// get the ReconciliationServiceItem from the ReconciliationService
+		$reconciliationServiceItem = $reconciliationService->getOrCreateItemByStatementUrl(
+			$propertyId,
+			$url
+		);
+
+		// assert that it matches the saved item
+		$this->assertSame( 'test label',
+			$reconciliationServiceItem->getItem()->getLabels()->getByLanguage( 'en' )->getText() );
+		$this->assertSame( $revisionId, $reconciliationServiceItem->getRevision() );
```
toban, line 89: We could add an assertion here that https://github.com/wmde/WikibaseReconcileEdit/pull/46/files#diff-8ece74e8b7d4a18586c072102e94f6da6d3bee0c117e6a71b6c5e865c63aba12R61 has happened also maybe?

Might be out of scope for this test class though.

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @covers \MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ReconciliationService
+ * @license GPL-2.0-or-later
+ */
+class ReconciliationServiceTest extends MediaWikiIntegrationTestCase {
+
+	public function testRemembersRevisionAfterReplaceWith(): void {
+		$reconciliationService = WikibaseReconcileEditServices::getReconciliationService();
+		$propertyId = new PropertyId( 'P1' );
+		$url = 'https://example.com/' . random_int( 0, PHP_INT_MAX );
```
toban, line 22: is this random_int required / doing anything useful here?

toban (commented):
Nice, some ideas that popped up when looking through it!

includes/MediaWiki/Api/BatchEditEndpoint.php
lucaswerkmeister: Good point.

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
lucaswerkmeister: I feel like that’s just a paranoia measure that shouldn’t be tested, to be honest. (I wasn’t even sure if I wanted to do it at all.)

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
lucaswerkmeister: Well, it’s ensuring that there’s no existing item with that URL, since we’re using real services.

includes/Reconciliation/ReconciliationServiceItem.php
lucaswerkmeister: Sounds okay. Should we go for `updateFromRevision`, or something similar? ^^ (I initially added just `setRevision( int $revisionId )`, but then realized the Item needed to be set as well, and at that point “replace” made sense to me, since both members of the object are being replaced.)

lucaswerkmeister: (Any thoughts about `ReconciledItem::finish()`, by the way? I was less sure about that command :D)

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
toban: fair enough

tests/phpunit/integration/Reconciliation/ReconciliationServiceTest.php
toban: gotcha

includes/Reconciliation/ReconciliationServiceItem.php
toban: updateFromRevision / setRevision both works for me :+1:

toban: Hmm, good question.. `finalize`? but might give the wrong ideas depending on who reads it. :D 

Can't really come up with anything better i think :)

lucaswerkmeister: I went with `updateFromEntityRevision()` so it doesn’t look so similar to `getRevision()` (maybe `getRevision()` should be `getRevisionId()` but I don’t want to do that here).

toban (approved):

lucaswerkmeister:
Rebased without changes.

PR #47, opened 2021-06-03
[T282236.2] Add special page

toban:
After https://github.com/wmde/WikibaseReconcileEdit/pull/47/commits/bd005dd61896c29e34039a2183200f7fcacd3de8 I decided it might make more sense to use javascript to issue this request rather than having the page submit the entire form.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -21,6 +21,19 @@
 			"i18n"
 		]
 	},
+	"SpecialPages": {
+		"WikibaseReconcileEdit": {
+			"class": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Specials\\SpecialWikibaseReconcileEdit",
+			"factory": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Specials\\SpecialWikibaseReconcileEdit::factory",
+			"services": [
+				"WikibaseReconcileEdit.EditRequestParser",
+				"WikibaseReconcileEdit.EditRequestSaver"
+			]
+		}
+	},
+	"ExtensionMessagesFiles": {
+		"WikibaseReconcileEditAlias": "WikibaseReconcileEdit.i18n.alias.php"
+	},
 	"RestRoutes": [
 		{
 			"path": "/wikibase-reconcile-edit/v0/edit",
--- a/includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
+++ b/includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
@@ -85,11 +85,9 @@ class SpecialWikibaseReconcileEdit extends SpecialPage {
 
 	public function execute( $subPage ) {
 		parent::execute( $subPage );
+
 		$this->requireLogin();
 		$this->showRequestForm();
-
-		$out = $this->getOutput();
-		$out->addModules( 'ext.WikibaseReconcileEdit' );
 	}
 
 	public function submitForm( $data ) {
--- a/includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
+++ b/includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
@@ -1,0 +1,165 @@
+<?php
+
+namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Specials;
+
+use HTMLForm;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequestParser;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequestSaver;
+use MediaWiki\Extension\WikibaseReconcileEdit\ReconciliationException;
+use MediaWiki\Message\Converter;
+use MediaWiki\Rest\LocalizedHttpException;
+use SpecialPage;
+use Status;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\Lib\Store\EntityRevision;
+use Wikibase\Lib\Store\EntityUrlLookup;
+use Wikibase\Repo\WikibaseRepo;
+
+/**
+ * @license GPL-2.0-or-later
+ */
+class SpecialWikibaseReconcileEdit extends SpecialPage {
+
+	/** @var string */
+	private $reconcileDefaultJSON;
+
+	/** @var string */
+	private	$entityDefaultJSON;
+
+	/** @var EditRequestParser */
+	private $editRequestParser;
+
+	/** @var EditRequestSaver */
+	private $editRequestSaver;
+
+	/** @var EntityUrlLookup */
+	private $entityUrlLookup;
+
+	public function __construct(
+		EditRequestParser $editRequestParser,
+		EditRequestSaver $editRequestSaver,
+		EntityUrlLookup $entityUrlLookup
+	) {
+		parent::__construct( 'WikibaseReconcileEdit' );
+
+		$this->editRequestParser = $editRequestParser;
+		$this->editRequestSaver = $editRequestSaver;
+		$this->entityUrlLookup = $entityUrlLookup;
+
+		$propertyId = new PropertyId( 'P1' );
+
+		$this->reconcileDefaultJSON = json_encode( [
+			EditRequestParser::VERSION_KEY => '0.0.1',
+			'urlReconcile' => $propertyId->serialize(),
+		], JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES );
+
+		$this->entityDefaultJSON = json_encode( [
+			EditRequestParser::VERSION_KEY => '0.0.1/minimal',
+			'statements' => [
+				[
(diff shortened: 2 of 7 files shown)


guergana:
Works like a charm. :heart_eyes: I give it a +1 because maybe @lucaswerkmeister also wants to take a look?

guergana (dismissed):
Works great!

extension.json
```diff
@@ -21,6 +21,26 @@
 			"i18n"
 		]
 	},
+	"SpecialPages": {
+		"WikibaseReconcileEdit": {
+			"class": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Specials\\SpecialWikibaseReconcileEdit",
+			"factory": "MediaWiki\\Extension\\WikibaseReconcileEdit\\MediaWiki\\Specials\\SpecialWikibaseReconcileEdit::factory"
+		}
+	},
+	"ResourceModules": {
```
lucaswerkmeister: I assume the JS code can be dropped now that the special page handles form submissions?

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+	private $editRequestParser;
+
+	/** @var EditRequestSaver */
+	private $editRequestSaver;
+
+	public function __construct() {
+		parent::__construct( 'WikibaseReconcileEdit' );
+
+		$this->editRequestParser = WikibaseReconcileEditServices::getEditRequestParser();
+		$this->editRequestSaver = WikibaseReconcileEditServices::getEditRequestSaver();
```
lucaswerkmeister: These two can be injected by specifying their service names in `extension.json`.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+			);
+		} catch ( LocalizedHttpException $exception ) {
+			return Status::newFatal( \Message::newFromKey( $exception->getMessageValue()->getKey() ) );
+		}
+
+		/** @var EntityRevision $entityRevision */
+		$entityRevision = $status->getValue()[0]['revision'];
+
+		$entityUrl = WikibaseRepo::getDefaultInstance()
+			->getEntityUrlLookup()
```
lucaswerkmeister: This should ideally be injected into the constructor (by the factory method, I guess), rather than being created ad-hoc here.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+			->getLinkUrl( $entityRevision->getEntity()->getId() );
```
lucaswerkmeister: I suspect this should be `getFullUrl()`, not `getLinkUrl()` – looking at some [redirects in codesearch](https://codesearch.wmcloud.org/search/?q=%5C%24this-%3EgetOutput%5C(%5C)-%3Eredirect&i=nope&files=&excludeFiles=&repos=), most of the URLs seem to come from `getFullURL()`, and only a handful (none of them in MediaWiki core) from `getLinkURL()`.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+			'required' => true,
+			'default' => $this->reconcileDefaultJSON,
+			'name' => 'reconcile'
+
+		];
+
+		$fields['entity'] = [
+			'type' => 'textarea',
+			'label-message' => 'wikibasereconcileedit-special-entity-label',
+			'maxlength' => 255,
```
lucaswerkmeister: Hm, this feels pretty short, to be honest… do we need to limit it this much?

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+
+		try {
+			$request = $this->editRequestParser->parseRequestBody( $data );
+			$status = $this->editRequestSaver->persistEdits(
+				[ $request ],
+				$data['token'],
+				$this->getContext()->getUser()
+			);
+		} catch ( LocalizedHttpException $exception ) {
+			return Status::newFatal( \Message::newFromKey( $exception->getMessageValue()->getKey() ) );
```
lucaswerkmeister: I think we can use `$this->msg()` instead of `\Message::newFromKey()`?

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,149 @@
...
+	public function execute( $subPage ) {
+		parent::execute( $subPage );
+		$this->requireLogin();
+		$this->showRequestForm();
+
+		$out = $this->getOutput();
+		$out->addModules( 'ext.WikibaseReconcileEdit' );
+	}
+
+	public function sendRequest( $data ) {
```
lucaswerkmeister: I suggest renaming this, since I wouldn’t say it sends a request now. Looking at [`setSubmitCallback` codesearch results](https://codesearch.wmcloud.org/search/?q=setSubmitCallback%5C(%20%5C%5B%20%5C%24this&i=nope&files=&excludeFiles=&repos= ), `onSubmit` or `submitForm` seem to be popular choices, and I think either of those would be fine – it should be okay to not mention in the method name _what_ the submit does, since there’s only one submit method anyways.

lucaswerkmeister:
Rebased onto #52.

extension.json
toban: I was thinking this could be the "rich" experience we give users that have javascript enabled.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: good idea

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: done

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: done

extension.json
lucaswerkmeister: In that case, I have to point out that `response.html(JSON.stringify(res, null, 2))` is a security vulnerability. Also, we should set up eslint to ensure consistent code style.

I suggest at least deferring this to a separate commit/PR, since the special page should work just fine without it.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: cool :+1:

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: okey, done

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
toban: done

toban:
> yeah it seemed to work! (but without parameters maybe, gonna try that)

I'm not so sure, need to try this further and should also catch ReconciliationException for example

extension.json
toban: broke them up into separate patches, this was also rebased on top of T282236.1.

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,165 @@
...
+		return 'wikibase';
+	}
+
+	public function execute( $subPage ) {
+		parent::execute( $subPage );
+		$this->requireLogin();
+		$this->showRequestForm();
+
+		$out = $this->getOutput();
+		$out->addModules( 'ext.WikibaseReconcileEdit' );
```
lucaswerkmeister: doesn’t exist in this commit

includes/MediaWiki/Specials/SpecialWikibaseReconcileEdit.php
```diff
@@ -0,0 +1,165 @@
...
+			);
+		}
+
+		/** @var EntityRevision $entityRevision */
+		$entityRevision = $status->getValue()[0]['revision'];
+
+		$entityUrl = $this->entityUrlLookup
+			->getFullUrl( $entityRevision->getEntity()->getId() );
+
+		if ( $entityUrl ) {
```
lucaswerkmeister, line 116: And what happens otherwise?

toban: It would just return to the form.

PR #48, opened 2021-06-03
Move EditEndpoint.js -> SingleEditEndpoint.js

toban:
There is no tests for the baseclass


guergana (approved):
Seems fair. Good catch.

PR #49, opened 2021-06-03
Catch PropertyDatatypeException

toban:
- adds a test for this


lucaswerkmeister (commented):
Does this really belong in the JS API integration tests? I think the test should be in the MinimalItemInput unit test – this kind of error handling isn’t a feature that needs to be tested at such a high level, in my opinion. (Also, it’s easier to test that the right message is used at the PHP level.)

lucaswerkmeister (approved):
I still don’t think this should have an API integration test, but okay.

PR #50, opened 2021-06-03
Add any errors that come from Status to the response

toban:
- batch endpoint
- single endpoint
- adds unit tests for EditEndpoint

### DIFF
--- a/includes/MediaWiki/Api/BatchEditEndpoint.php
+++ b/includes/MediaWiki/Api/BatchEditEndpoint.php
@@ -77,7 +77,7 @@ class BatchEditEndpoint extends EditEndpoint {
 
 		$response = $this->getResponseBody( $status );
 
-		if ( $status->isGood() ) {
+		if ( $status->isOK() ) {
 			$response['results'] = array_map( static function ( $result ) {
 				/** @var EntityRevision $entityRevision */
 				$entityRevision = $result['revision'];
--- a/includes/MediaWiki/Api/BatchEditEndpoint.php
+++ b/includes/MediaWiki/Api/BatchEditEndpoint.php
@@ -68,17 +68,16 @@ class BatchEditEndpoint extends EditEndpoint {
 			$otherItems = $request->otherItems();
 
 			$otherStatus = $this->persistItem( $reconciledItem, $otherItems );
+			$status->merge( $otherStatus );
 			if ( !$otherStatus->isOk() ) {
 				break;
 			}
 			$status->setResult( true, array_merge( $status->getValue(), [ $otherStatus->getValue() ] ) );
-			$status->merge( $otherStatus );
 		}
 
-		$response = [
-			'success' => $status->isOK(),
-		];
-		if ( $status->isOK() ) {
+		$response = $this->getResponseBody( $status );
+
+		if ( $status->isGood() ) {
 			$response['results'] = array_map( static function ( $result ) {
 				/** @var EntityRevision $entityRevision */
 				$entityRevision = $result['revision'];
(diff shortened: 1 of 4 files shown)


includes/MediaWiki/Api/BatchEditEndpoint.php
```diff
@@ -68,17 +68,16 @@ public function run() {
...
-			$status->merge( $otherStatus );
 		}
 
-		$response = [
-			'success' => $status->isOK(),
-		];
-		if ( $status->isOK() ) {
+		$response = $this->getResponseBody( $status );
+
+		if ( $status->isGood() ) {
```
lucaswerkmeister: I’m not convinced by the change from `isOK()` to `isGood()` here… if the status contains a result, doesn’t it make sense to add it to the response, even if there are also errors?

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -141,6 +142,31 @@ public function needsWriteAccess() {
...
 	}
 
+	/**
+	 * Returns the base for the response body
+	 * @param Status $status
+	 * @return array
+	 */
+	public function getResponseBody( Status $status ): array {
+		$response = [
+			'success' => $status->isGood(),
```
lucaswerkmeister, line 152: I’m also not sure about the use of `isGood()` vs. `isOK()` here, and feel like `isOK()` might make more sense… if users of the API want to check if there were errors or not, they can look at the `errors` array, no?

includes/MediaWiki/Api/EditEndpoint.php
```diff
@@ -141,6 +142,31 @@ public function needsWriteAccess() {
...
+			'errors' => []
+		];
+
+		$errors = $status->getErrors();
+		if ( !empty( $errors ) ) {
+			$response['errors'] = array_map( function ( $errorMessage ) {
+				$msg = ApiMessage::create( $errorMessage );
```
lucaswerkmeister, line 159: I’m a bit uncomfortable with the use of `ApiMessage` here, since as far as I can tell the REST API otherwise wants to use `MessageValue` instead of `Message`… but `Message` is what we get from the `Status`, and the MediaWiki core class to convert between the two (`Converter`) isn’t stable, so this is probably the best we can do.

includes/MediaWiki/Api/BatchEditEndpoint.php
toban: yeah this makes sense.

includes/MediaWiki/Api/EditEndpoint.php
toban: To some degree I can agree with this. But feels like we are hiding something from the API consumer, personally I would like to get notified in as many ways as possible if the request completed but might have done something fishy.

To just match the term `success` to what fits the description of these functions I'd also go with `isGood` 

```
IsGood - the operation completed and didn't have any error or warnings
isOK - whether the operation completed
```

includes/MediaWiki/Api/EditEndpoint.php
toban: Yeah I wasn't entirely sure what is supposed to format/handle these errors either. The REST API seems to also use these but then converting them to message [when throwing localized HTTP errors](https://github.com/wikimedia/mediawiki/blob/REL1_35/includes/Rest/Handler/ActionModuleBasedHandler.php#L138).

includes/MediaWiki/Api/EditEndpoint.php
lucaswerkmeister: Hm, fair enough. I guess the success field is redundant either way – `isOK()` is (now) reflected by whether there’s an entity in the response, and `isGood()` by whether there are any errors. So maybe `isGood()` is better for the `success` field.

(I don’t think we should have two fields in the response, for good and OK – I think this confusion should stay within the extension code, not be exposed to API users as well :D )

lucaswerkmeister (approved):

PR #51, opened 2021-06-03
Update extension description message

lucaswerkmeister:
I tried to make it useful without tying it too much to the current reconciliation mechanism (exact match on one URL statement), since later versions of the extension might support more reconciliation options.

Bug: [T284224](https://phabricator.wikimedia.org/T284224)


toban:
Cool, good catch :D

PR #52, opened 2021-06-04
Add EditRequestSaver

toban:
- move saving of edit requests to its own service

### DIFF
--- a/includes/MediaWiki/Api/SingleEditEndpoint.php
+++ b/includes/MediaWiki/Api/SingleEditEndpoint.php
@@ -3,7 +3,7 @@
 namespace MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Api;
 
 use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequestParser;
-use MediaWiki\Extension\WikibaseReconcileEdit\Reconciliation\ItemReconciler;
+use MediaWiki\Extension\WikibaseReconcileEdit\MediaWiki\Request\EditRequestSaver;
 use MediaWiki\Extension\WikibaseReconcileEdit\ReconciliationException;
 use MediaWiki\Rest\LocalizedHttpException;
 use MediaWiki\Rest\Validator\JsonBodyValidator;
@@ -11,8 +11,6 @@ use MediaWiki\Session\SessionProviderInterface;
 use RequestContext;
 use User;
 use Wikibase\Lib\Store\EntityRevision;
-use Wikibase\Repo\EditEntity\MediawikiEditEntityFactory;
-use Wikibase\Repo\WikibaseRepo;
 use Wikimedia\ParamValidator\ParamValidator;
 
 /**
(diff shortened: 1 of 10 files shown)


includes/MediaWiki/Api/SingleEditEndpoint.php
```diff
@@ -59,29 +50,15 @@ public function run() {
...
+		$saveStatus = $this->editRequestSaver->persistEdits( [ $request ], $this->editToken, $this->user );
 
 		// Make some sort of response
 		$response = [
 			'success' => $saveStatus->isGood()
 		];
 		if ( $saveStatus->isGood() ) {
 			/** @var EntityRevision $entityRevision */
-			$entityRevision = $saveStatus->getValue()['revision'];
+			$entityRevision = $saveStatus->getValue()[0]['revision'];
```
toban, line 59: this might need a check for that element maybe?

toban: Or maybe not, if we reach this point and there is nothing here It should crash and burn probably.

lucaswerkmeister: :+1:

includes/MediaWiki/Request/EditRequestSaver.php
```diff
@@ -0,0 +1,124 @@
...
+				$request->reconcilePropertyId()
+			);
+			$otherItems = $request->otherItems();
+
+			$otherStatus = $this->persistItem( $reconciledItem, $otherItems, $editToken, $user );
+			if ( !$otherStatus->isOk() ) {
+				break;
+			}
+			$status->setResult( true, array_merge( $status->getValue(), [ $otherStatus->getValue() ] ) );
+			$status->merge( $otherStatus );
```
lucaswerkmeister: This line should move up, I think, like it did in #50?

includes/MediaWiki/Request/EditRequestSaver.php
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		return $status;
+	}
+
+	/**
+	 * @param ReconciledItem $reconciledItem
+	 * @param ReconciliationServiceItem[] $otherItems
```
lucaswerkmeister, line 64: The `$editToken` and `$user` params seem to be missing.

includes/ServiceWiring.php
```diff
@@ -29,6 +30,18 @@
 		);
 	},
 
+	'WikibaseReconcileEdit.EditRequestSaver' => function ( MediaWikiServices $services ): EditRequestSaver {
+		$repo = WikibaseRepo::getDefaultInstance();
+		$editEntityFactory = method_exists( $repo, 'getEditEntityFactory' )
+			? $repo->getEditEntityFactory() // 1.36+
```
lucaswerkmeister: Hm, in the 1.36 version we could actually pass `$services` into this method, to ensure that the edit entity factory comes from the same service container. But I’m not sure if that’s very important here.

lucaswerkmeister (commented):
Looks alright so far, but there are no tests for the new class yet…

includes/MediaWiki/Request/EditRequestSaver.php
toban: yup, good catch, must have moved back when rebasing.

includes/MediaWiki/Request/EditRequestSaver.php
toban: done

includes/ServiceWiring.php
toban: okay

lucaswerkmeister (approved):

PR #56, opened 2021-08-10
add virtual machine to inventory

dati18:


toban:
We used this to debug https://phabricator.wikimedia.org/T286292

Seems to be mostly working for setting up the system, however when started the server does not respond on 80 or redirects incorrectly to https, will not spend more time looking into it now.

PR #57, opened 2021-08-16
add virtual machine to inventory

dati18:


toban:
Seems composer wasn't happy about the php version, bump to the new mediawiki image and removed the version specified for gettext-base.

It works now :1st_place_medal:

toban (approved):

